PR #3, opened 2017-06-20
Update modules.js

locks:
- Use destructuring for `params` argument
- Use template string for interpolation


steveklabnik:
Fancy, thanks!

PR #25, opened 2017-07-14
Remove clap as a dependency from the Library

### LINKED ISSUE #20: Remove library's clap dependency
https://github.com/steveklabnik/rustdoc/pull/19 fixed https://github.com/steveklabnik/rustdoc/issues/14, but since `Config::new` uses `clap` in its arguments, the library has to depend on `clap` as well, which isn't as clean as I'd like.

Removing this dependency is a good thing, as only `main` should need to depend on `clap`.

mgattozzi:
We only want to have the main binary to have the clap crate imported.
This is because it's a command line arg parsing crate. Ideally we would
pass values from it to methods as needed and not pass clap's data types
functions outside of main.

This commit does just that by removing clap from lib.rs and generating
a PathBuf for the Config struct in main.rs instead.

Closes #20

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,7 +27,6 @@ struct Asset {
 
 impl Config {
     pub fn new(manifest_path: PathBuf) -> Result<Config, Box<std::error::Error>> {
-        // unwrap is okay because we take a default value
         let host = generate_analysis(&manifest_path)?;
 
         let assets = vec![Asset {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,7 +1,6 @@
 extern crate jsonapi;
 extern crate rls_analysis as analysis;
 extern crate serde_json;
-extern crate clap;
 
 use analysis::raw::DefKind;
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -27,9 +26,8 @@ struct Asset {
 }
 
 impl Config {
-    pub fn new(matches: &clap::ArgMatches) -> Result<Config, Box<std::error::Error>> {
+    pub fn new(manifest_path: PathBuf) -> Result<Config, Box<std::error::Error>> {
         // unwrap is okay because we take a default value
```
aochagavia: This comment seems to be no longer necessary

mgattozzi: Yeah you're definitely right. I forgot to grab that when I moved it over.

steveklabnik:
Awesome, thanks so much!

mgattozzi:
You're welcome! I'll be putting some more in today hopefully.

PR #26, opened 2017-07-14
Replace string formatting by path functions

### LINKED ISSUE #9: Improve filesystem handling
Right now, [this](https://github.com/steveklabnik/rustdoc/blob/99685f23ea55af4dc641c953c1f9fc5df8235b46/src/main.rs#L155-L156) code uses string concatenation to build a file path. We should instead use the actual `Path` APIs.

### LINKED ISSUE #10: Improve paths with Cargo integration
[This code](https://github.com/steveklabnik/rustdoc/blob/99685f23ea55af4dc641c953c1f9fc5df8235b46/src/main.rs#L294-L300) uses string concatenation instead of real `Path` APIs. We should fix this.

aochagavia:
Closes #9 and closes #10

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -123,8 +123,7 @@ pub fn build(config: &Config) -> Result<(), Box<std::error::Error>> {
         }
     }
 
-    // TODO: use real fs handling here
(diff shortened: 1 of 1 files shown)


steveklabnik:
Great, thanks so much!

PR #28, opened 2017-07-14
Improve travis-ci with rustfmt and eslint

### LINKED ISSUE #17: Improve CI setup
In https://github.com/steveklabnik/rustdoc/issues/12 we added Travis. It only currently runs `cargo test`/`npm run test`.

We should also:

- Run rustfmt, and fail if its output isn't correct
- Run `npm run eslint` and fail if it fails

mgattozzi:
Currently we had no lint to check that someone had run rustfmt on the
code. This makes sure that someone runs rustfmt so they can pass the CI
checks now. We also wanted eslint to run but according to the
conversation in issue #17 this is already the case when tests are run
using the ember-cli tool.

This adds a ci-scripts folder that we can fill with more scripts we need
run during testing while not cluttering up travis with all kinds of
conditionals and things. In this case we add the rustfmt.sh script which
checks if rustfmt is already installed from the cache and if not to
install it. If it's out of date compared to upstream it'll install the
newest version of rustfmt automatically then cache it for later!

This commit also reformats lib.rs as it was not compliant with rustfmt
and was causing the CI to fail under the new changes

Closes #17

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,14 @@ rust:
   # - beta
   - nightly
 
+cache:
+  - cargo
+
 before_install:
   - nvm install 8.1.4
+  - ./ci-scripts/rustfmt.sh
 
 script:
   - cargo test
+  - rustfmt --write-mode diff src/*
   - cd frontend && npm install && npm run test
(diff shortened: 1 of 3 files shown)


steveklabnik:
This looks great, thanks a ton! The `lib.rs` bit shows why this is so important; I try to keep up with it, but even then I ended up failing!

let's see these pass on travis and then I'm down to merge!

mgattozzi:
No problem! Hopefully this will catch these problems sooner rather than later and we can add more lints over time for other things that should be enforced in the code base.

PR #29, opened 2017-07-14
Add ability to only output JSON

### LINKED ISSUE #27: Add a flag to only produce the JSON
Right now, `rustdoc build` produces HTML, CSS, JS, and JSON. Adding a flag to only produce the JSON is a good idea.

mgattozzi:
Currently rustdoc can only output everything needed. We want it to also
only output just the JSON so that others can use it's values to do other
things if they want it. Adds this functionality by adding a flag to
build. The command `rustdoc build -j` or `rustdoc build --json` will
only output JSON now.

Close #27 

@steveklabnik I wasn't sure what you had in mind and this is a preliminary stab at it. currently it just outputs still to `target/doc` with just the JSON. Let me know how you would want this to really work/what can be improved.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,6 +15,15 @@ pub struct Config {
     manifest_path: PathBuf,
     host: analysis::AnalysisHost,
     assets: Vec<Asset>,
+    build_output: BuildOutput,
+}
+
+/// BuildOutput is used to determine what `rustdoc` should output when building.
+/// Currently it either builds and outputs just JSON or HTML/CSS/JS and JSON
+#[derive(PartialEq, Eq)]
+pub enum BuildOutput {
+    Json,
+    All,
 }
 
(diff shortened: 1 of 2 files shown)


euclio:
Whoops, I just finished an alternative implementation of this. My idea was to have an `emit` flag like rustc, where one can supply `--emit=json,assets` or any combination. I can submit what I have too if it's helpful.

steveklabnik:
Looks good to me, though yeah, we should consider if @euclio 's approach is better!

I'm leaning towards "yes" though it is early on a monday; @mgattozzi any thoughts?

mgattozzi:
I'm thinking that @euclio's approach will be better long term and more flexible then this approach. I'm thinking that would be best to do it that way! @steveklabnik I'm going to close this PR in favor of @euclio's approach.

PR #35, opened 2017-07-16
Give a bit more context

GuillaumeGomez:


steveklabnik:
Hey @GuillaumeGomez ! I like this, but it fails `rustfmt`, could you fix this please?

GuillaumeGomez:
```
-        CrateErr {⏎
-            error: format!("Crate not found: \"{}\"", crate_name),⏎
-        }⏎
+        CrateErr { error: format!("Crate not found: \"{}\"", crate_name) }
```

Seriously? ><

mgattozzi:
As an aside might be worth spending sometime specifying the options for formatting w/ rustfmt. For instance I prefer 80 char length lines as it makes it easier to read online but rustfmt defaults to 100. Might be worth discussing in an issue what to enforce as the style with rustfmt, but it's not really pressing.

mgattozzi:
To each there own! I'm with you on this one though. I prefer the braces on separate lines for structs if possible

GuillaumeGomez:
I generally don't like rustfmt since it makes the code generally more ugly... But whatever. ;)

steveklabnik:
In general, the idea is that `rust-lang` projects will be using `rustfmt` without changing the defaults; there will be some change as `rustfmt` isn't at 1.0 yet, but I'd like to stick with it here too.

Thanks!

PR #36, opened 2017-07-17
Remove hardcoding of Assets from rustdoc

### LINKED ISSUE #8: Fix the "hilarious hard-coding"
Right now, the hash of the frontend files is coded into the binary, so when we update the front end, it won't get picked up automatically.

To fix:

1. Instead of hardcoding, have it read the directory to determine the file names.
2. Modify the `Assets` struct to be able to handle a name as well as the contents. Possibly, this should instead be an `Asset` struct with name + contents, and the `Assets` struct could be a `Vec<Asset>`.
3. use these names when writing out the files.

* https://github.com/steveklabnik/rustdoc/blob/99685f23ea55af4dc641c953c1f9fc5df8235b46/src/main.rs#L42
* https://github.com/steveklabnik/rustdoc/blob/99685f23ea55af4dc641c953c1f9fc5df8235b46/src/main.rs#L249

### LINKED ISSUE #18: Amazing CI setup
https://github.com/steveklabnik/rustdoc/issues/17 gives us good CI, this ticket is for *great* CI.

We should write a custom lint that checks if anything in `frontend` was modified, and makes sure that the `frontend\dist` folder was also modified. If not, it should fail the build.

mgattozzi:
This adds a build script to dynamically generate the assets vector used
in the creation of a new config. It does so by finding all of the assets
and creating an expression in src/assets.in. This expression looks much
like what the `Config` implementation had originally, mostly using
a name and `include_str!()`. Now though we don't have to worry about
changing the names every time as a result!

A few changes have also been made to the CI:

- Namely it tests that it can run on the example crate after having been
  run
- Makes sure rustfmt is run on the build.rs script
- Makes sure rustfmt is run on all .rs files in src only
- Dist files are generated each time to make sure it works dynamically
- Frontend tests are run first to make sure that the back will have
  files to include later on

While this still means the files are tracked and we still depend on
ember the crate can still be compiled as is until a better solution is
found. This also generates the latest assets available and checks them
into the source tree.

Closes #8
Closes #18


build.rs
```diff
@@ -0,0 +1,116 @@
...
+use std::fmt;
+use std::path::Path;
+use std::fs::{File, read_dir, metadata};
+use std::io::{Write, BufWriter};
+
+// Location of the dist folder
+#[cfg(not(target_os = "windows"))]
+const DIST: &str = "frontend/dist/";
+#[cfg(target_os = "windows")]
+const DIST: &str = "frontend\\dist\\";
```
killercup: `r"Raw\strings\are\your\friend"` :) (Can Windows _really_ not deal with forward slashes here?)

build.rs
```diff
@@ -0,0 +1,116 @@
...
+/// Intermediary data structure used to hold the path of the file
+pub struct Asset {
+    pub path: String,
+}
+
+// This gives us the ToString impl as well. We're using this to write out what
+// an asset would look like for the actual program into the asset.in file.
+impl fmt::Display for Asset {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut buffer = String::from("Asset { name: \"");
```
killercup: Have a look at https://doc.rust-lang.org/nightly/std/fmt/struct.Formatter.html#method.debug_struct and its friends :)

build.rs
mgattozzi: I wasn't sure given the nature of that's how it normally does paths. I didn't even know raw strings were a thing. I knew about byte ones

build.rs
mgattozzi: I will see if something here can really help out!

build.rs
steveklabnik: windows can deal with the `\`s. The only downside is if there's an error message

    C:\it\looks\like/this/which/is/weird

but it's fine

build.rs
```diff
@@ -0,0 +1,116 @@
...
+        vec![
+            "frontend\\dist\\ember-fetch"
+        ]
+    };
+}
+#[cfg(not(target_os = "windows"))]
+lazy_static! {
+    static ref IGNORE: Vec<&'static str> = {
+        vec![
+            "frontend/dist/ember-fetch"
```
steveklabnik: so, I'm actually not entirely sure we want to ignore this file; I recently inquired as to why it's there, and it's for fastboot. I don't think we'd ever want to require fastboot, but we might want to give people the option? We can of course punt this question to later.

build.rs
```diff
@@ -0,0 +1,116 @@
...
+
+/// Start the recursion from the top level of the frontend's dist folder
+fn acquire_assets() -> Vec<Asset> {
+    recurse_dir(&Path::new(DIST))
+}
+
+/// Recurse through a directory, if it's an asset file add it to our list and if
+/// it's a directory get all the assets in there then put the `Assets` returned
+/// into this directory's Vector and return that.
+fn recurse_dir<P>(path: &P) -> Vec<Asset>
```
steveklabnik: what about the `walkdir` crate here?

build.rs
steveklabnik: also given that there's only one of them, might be easier to not use a `Vec` and not need to pull in lazy-static, etc

build.rs
mgattozzi: Okay I can remove that then!

build.rs
mgattozzi: It seemed to fail when it was included for some reason, but I can double check. My thinking with the `Vec` was if over time we needed to ignore files it would be there, but it's fairly easy to remove.

build.rs
mgattozzi: I didn't even know that was a thing and would be more ergonomic then this solution, at least from a maintenance perspective

build.rs
mgattozzi: @steveklabnik I've confirmed that the ember dir causes the whole thing to fail documenting code in the example crate though I'm unsure why. We might want to keep this available for now (until that issue is fixed somehow) or just have it check against that path.

steveklabnik: cool, please file a bug about it after we merge!

build.rs
steveklabnik: yeah it also handles tricky things like "recurse so many times you blow the stack" and such 😄

mgattozzi:
I've made some updates and actually removed the hard coded files from the repo which I forgot to before. I would like to squash and merge all of this once it's done. I've also updated the comment above to also close #18 when this is merged based off the conversation there.

.gitignore
```diff
@@ -1,2 +1,4 @@
 target/
 **/*.rs.bk
+src/asset.in
+frontend/dist/
```
steveklabnik: This now means we require `ember` to build `rustdoc`, right?

That's something I was trying to get away from...

mgattozzi: Yes that would be the case unfortunately. However, I'm not really sure how this would be done without that. Dynamic means we need whatever creates it unfortunately. Otherwise we'd have to still update the assets manually. Which is possible but kind of a pain. :/ I'm not really sure what the best course of action would be.

mgattozzi: Maybe for now we have the dependency then make it work in a more pluggable way for when people use it? I think this is what you wanted to do eventually correct?

build.rs
mgattozzi: Will do!

.gitignore
steveklabnik: I think updating the assets manually is fine, it's the "needing to copy/paste the new hashes into the source code" that's the real issue. 

Hm.

mgattozzi: If we update the assets and they're version controlled then the build script will work as it is now. It's not dependent on ember but it's still inflexible in a way. How about for now keeping the frontend/dist folder checked into git and then having the build script. I think that's a good medium until a more pluggable idea can be come up with.

steveklabnik: @mgattozzi sounds good to me!

mgattozzi: Okay I'll merge in the latest changes and fix this up and we should be good to merge. I'll open up separate issues for things that arose in this PR.

mgattozzi:
I had to rebase and some other things to get this working and cleaned up but we should be good to go. I've updated the top level comment with the newest commit message as there were some things that have changed during review.

mgattozzi:
Granted I think I could get the dist files built out of the commit if necessary. It was just a hairy mess with git getting it to this point like you wouldn't believe.

steveklabnik:
So, this looks great to me! One problem:

```text
error: unknown character escape: f
 --> src\asset.in:3:15
  |
3 | name: "assets\frontend-34abb755950f67fc6df73532cf4bd305.js",
  |               ^

error: unknown character escape: f
 --> src\asset.in:4:49
  |
4 | contents: include_str!("../frontend/dist/assets\frontend-34abb755950f67fc6df73532cf4bd305.js")
  |                                                 ^
```

😅  Let me see...

build.rs
```diff
@@ -0,0 +1,87 @@
...
+
+        // We know this will give the proper name so we can do this to get the
+        // proper name for the asset without the rest of the beginning path
+        buffer.push_str(
+            self.path
+                .split(DIST)
+                .skip(1)
+                .collect::<Vec<&str>>()
+                .pop()
+                .unwrap(),
```
steveklabnik: i guess in theory `str::replace` could be used to replace `\` with `/` here?

mgattozzi: Huh windows is weird. Should probably set up appveyor at some point as well. we should be able to use that

steveklabnik: agreed, maybe i'll set that up right now

mgattozzi:
@steveklabnik looks like it failed since appveyor got turned on but there's no appveyor file yet in the tree.

steveklabnik:
@mgattozzi lol, im working on one right now, sorta annoying that it did a build before i got there...

steveklabnik:
@mgattozzi looks like there's a rustfmt failure too

mgattozzi:
@steveklabnik hahaha well if it works for you locally then we can merge this. I was going to do all the error-chain stuff after this. It'll help clean up the build script too. I've squashed and put in the rustfmt fix. We should be good *fingers crossed*

steveklabnik:
It does work for me, but I'm gonna let appveyor build one last time before merging.

Thanks again for tackling this!

mgattozzi:
@steveklabnik no problem. Looks like replace didn't fix all of them so let me try this again.

steveklabnik:
No worries, I'm about to take off for most of the evening, so no rush 😄

mgattozzi:
This should do it. I tested locally on my windows machine. Fixed it by replacing it when the path name is originally found!

mgattozzi:
@steveklabnik Wooo finally done!

steveklabnik:
🤘 ❤️

Issue #11, opened 2017-07-12
Refactor flow of the backend


steveklabnik:
https://github.com/steveklabnik/rustdoc#the-backend right now is a lie, and everything is a mess. This should be fixed.

steveklabnik:
I am bumping this back to the backend milestone; this actually might be a nice "first contributor" issue.

jbcden:
@steveklabnik I would like to take a stab at this! I've looked through the project a little bit and was wondering if you had any suggestions for where to start.

steveklabnik:
@jbcden hey hey!

So basically it's like "make the code look like the README says", which involves untangling some fairly gnarly dependencies.

One thing at the start: because `Config` requires an `AnalysisHost` it currently [generates one in its constructor](https://github.com/steveklabnik/rustdoc/blob/68238a5bcb766291198a9aea34f8d326e3952dd3/src/lib.rs#L30). To break this dependency, it should
probably take the host as an argument to its constructor, that way we can generate the analysis first, and then put it in the `Config`.

However, since `build` takes a `Config`, and that feels right to me, that'd be weird, as now we'd need to call `generate_analysis` outside of `build`, and `build` should be the entry point, and take a `Config`. So from there, I'm thinking that instead, it's `generate_analysis` that's doing too much: it both creates an `AnalysisHost` *and* loads up the data. So instead, `Config`'s constructor should probably [have this line in it instead](https://github.com/steveklabnik/rustdoc/blob/68238a5bcb766291198a9aea34f8d326e3952dd3/src/lib.rs#L255), and then `generate_analysis` would take the host as an argument instead.

That'd be step one, and a patch for just that would even be cool.

I haven't totally thought through all the implications of the rest of it yet, but basically, what I'd like is for `build` to look like

```rust
pub fn build(config: &Config) -> Result<(), Box<std::error::Error>> {
    generate_analysis(config)?;
    
    let ast = DocAst::new(config)?;
    let json = ast.to_json()?;
    
    write_assets(config)?

    Ok(())
}
```

That is, the flow of it looks roughly like the flow in the README. This is tough because `DocAst` isn't a thing that exists yet... for now, it would require creating something that can handle Crates and Modules.

So, how about this: @jbcden maybe tackling that first bit is a good idea, and I'll noodle on the second half?

jbcden:
@steveklabnik Sounds like a plan! I'll try to get something together in the next couple of days.

steveklabnik:
[...]

PR #41, opened 2017-07-17
Inject config into generate_analysis

jbcden:
This is the first steps towards addressing https://github.com/steveklabnik/rustdoc/issues/11.


jbcden:
@steveklabnik I'm not sure if this goes a little further than you were envisioning. I ended up passing the entire `Config` into `generate_analysis` since the passing in the host would have meant passing two of `Config`'s three fields into `generate_analysis` separately. Let me know what you think :)

steveklabnik:
Looks great! Thanks a ton 😄

PR #42, opened 2017-07-18
Look for package name in manifest path

### LINKED ISSUE #30: rustdoc only works with example crate
It looks like the code is hard-coded to look for the example crate: https://github.com/steveklabnik/rustdoc/blob/ce9ee35612fa335b16e6cb4690b779d8b433743a/src/lib.rs#L95

If I run something like `cargo run -- --manifest-path=../hyper build` I get a panic when it tries to unwrap the iterator result.

I would like to help fix this, but would like to understand what should be done before writing some code. From what I gather, the intention is to look for the name of the root crate. Should we be looking at the `Cargo.toml` file to determine the root crate?

hjr3:
Use `cargo metadata` output to determine the root package name.

Fixes #30

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,8 +77,8 @@ pub fn generate_json(config: &Config) -> Result<JsonApiDocument, Box<std::error:
     "compiler_builtins"
     */
 
-    // FIXME: this whole code shouldn't look for a precise crate.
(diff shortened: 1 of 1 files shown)


hjr3:
Updated with feedback to use a shell command instead of using `cargo` as a dependency.

steveklabnik:
Awesome, thanks!

PR #47, opened 2017-07-18
Refactor flow of the backend

### LINKED ISSUE #11: Refactor flow of the backend

steveklabnik:
This ends up removing generate_json into better parts, and refactors
build to use those parts properly.

The new DocData struct reifies the previous 'data' hash that was in
generate_json, and gives it some structure.

DocData::to_json is still messy and won't work for arbitrary stuff
yet; that's the next big work to do!

I should note that originally, I conceived of `DocData` as an AST of sorts, but the problem with that is that for it all to be valid, you have to bring the entire tree up to build it. As such, the HashMap with linked IDs approach means that we don't have to; we can create the bits before they're done. This is important for really large docs trees, as we don't want to store literally everything in memory at all times, and also paves the way for things like streaming output.

Given that the `JsonAPI` crate ends up needing to build the entire tree, and it's painful, we might need to switch away from it in the future, dunno.

Fixes #11

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,16 +29,15 @@ Here's how it does it:
    `target\rls`.
 2. It reads those save analysis files with the `rls-analysis` crate. As you may be able
    to guess from the name, this is pretty much why it exists!
-3. It goes through the processed information and generates a sort of "docs AST"; starting
-   at the root of your crate and branching from there.
-4. It converts this "docs AST" to JSON, more specifically, [JSON API](https://jsonapi.org).
+3. It goes through the processed information and turns it into a `DocData` struct that
+   contains the top-level crate information, and the relevant information for everything
+   in the crate.
+4. It converts this `DocData` to JSON, more specifically, [JSON API](https://jsonapi.org).
 5. It writes out this JSON to the `target\doc` directory of the crate that
    it's documenting.
 6. It writes out some HTML/CSS/JS from the frontend `target\doc` too.
 
-Well, this is [how it's going to
-work](https://github.com/steveklabnik/rustdoc/issues/11), anyway: the code
-isn't exactly super clean at the moment. More work to do!
+You can also request it to only write out some of this information through the `--emit` flag.
 
 ### The frontend
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -175,8 +175,8 @@ impl DocData {
     fn new(config: &Config) -> Result<DocData> {
         let roots = config.host.def_roots()?;
 
-        let package = package_name_from_manifest_path(&config.manifest_path)?; 
-        let id = roots.iter().find(|&&(_, ref name)| name == &package); 
+        let package = package_name_from_manifest_path(&config.manifest_path)?;
+        let id = roots.iter().find(|&&(_, ref name)| name == &package);
         let root_id = match id {
             Some(&(id, _)) => id,
             _ => return Err(ErrorKind::CrateErr("example").into()),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,7 +7,7 @@ extern crate error_chain;
 pub mod error;
 use error::*;
 
-use std::collections::{BTreeMap, HashMap};
+use std::collections::HashMap;
 use std::fs::{self, File};
 use std::io;
 use std::io::prelude::*;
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -29,16 +29,15 @@ Here's how it does it:
...
+   in the crate.
+4. It converts this `DocData` to JSON, more specifically, [JSON API](https://jsonapi.org).
 5. It writes out this JSON to the `target\doc` directory of the crate that
    it's documenting.
 6. It writes out some HTML/CSS/JS from the frontend `target\doc` too.
 
-Well, this is [how it's going to
-work](https://github.com/steveklabnik/rustdoc/issues/11), anyway: the code
-isn't exactly super clean at the moment. More work to do!
+You can also request it to only write out some of this information through the `--emit` flag.
```
steveklabnik: I tossed this in here while I was working on the `README`; it could have been part of #31 , and rather than make a separate PR, I'm tossing it in here.

src/lib.rs
```diff
@@ -303,3 +191,167 @@ fn generate_analysis(config: &Config) -> Result<(), Box<std::error::Error>> {
...
+
+        let data = DocData::build_data(config, root_id, &mut krate)?;
+
+        Ok(DocData { krate, data })
+    }
+
+    fn build_data(
+        config: &Config,
+        root_id: analysis::Id,
+        krate: &mut Crate,
```
steveklabnik: this is slightly awkward, but is needed since we actually figure out what goes in the crate once we try to process all of the data.

we could instead return everything, I guess... dunno

src/lib.rs
```diff
@@ -303,3 +191,167 @@ fn generate_analysis(config: &Config) -> Result<(), Box<std::error::Error>> {
...
+struct Crate {
+    id: analysis::Id,
+    name: String,
+    docs: String,
+    modules: Vec<String>,
+}
+
+#[derive(Debug)]
+enum Item {
+    Module { name: String, docs: String },
```
mgattozzi: This will be expanded over time I'm guessing based off the `DefKind` matching block above right?

steveklabnik: Yup, that's the idea!

src/lib.rs
```diff
@@ -303,3 +191,167 @@ fn generate_analysis(config: &Config) -> Result<(), Box<std::error::Error>> {
...
+        use jsonapi::api::*;
+
+        let root_def = config.host.get_def(self.krate.id)?;
+
+        let mut document = JsonApiDocument::default();
+
+        let mut map = HashMap::new();
+        map.insert(
+            String::from("docs"),
+            serde_json::Value::String(root_def.docs.clone()),
```
jbcden: Could we get this off of `self.krate`?

src/lib.rs
```diff
@@ -303,3 +191,167 @@ fn generate_analysis(config: &Config) -> Result<(), Box<std::error::Error>> {
...
+            }
+        }
+
+        Ok(data)
+    }
+
+    fn to_json(&self, config: &Config) -> Result<String, Box<std::error::Error>> {
+        use jsonapi::api::*;
+
+        let root_def = config.host.get_def(self.krate.id)?;
```
jbcden: I'm not sure if there is a reason to keep this but it seems we could replace every instance of `root_def` with `krate`. This would also remove the need to pass the `Config` into this function.

src/lib.rs
jbcden: I think the reason this feels awkward is because passing in a mutable struct like this seems to emphasize some greater role in the `build_data` function when its current use is minimal. Looking at the resulting `data` HashMap, I think it might be interesting to explore encoding this data directly into the `Crate` struct. I'm not sure if that would be pushing too much information onto the `Crate` though. I don't think that change would need to be in this PR but just throwing out some food for thought.

euclio:
I'm currently working on some tests. It would be helpful to have a function that takes a package name and an `AnalysisHost` that goes to JSON. I can tackle that in my own PR if you'd like, however.

steveklabnik:
@euclio sounds good!

I am going to merge this and open an issue about the comments @jbcden made.

Issue #22, opened 2017-07-13
`cargo doc` doesn't work in this crate


steveklabnik:
I'd like to generate docs for `rustdoc`, but it's blocked on https://github.com/rust-lang/cargo/issues/4283

aochagavia:
With that issue closed as WONTFIX, are there any alternatives to make this work on Windows?

steveklabnik:
@aochagavia that's a good question; I'm not actually sure.

It may also be something that we can't fix until this becomes the "official" rustodoc. Hrm.

steveklabnik:
Looks like this is not necessarily WONTFIX.

PR #48, opened 2017-07-18
Configure appveyor

steveklabnik:


.appveyor.yml
```diff
@@ -0,0 +1,13 @@
...
+  - SET PATH=%PATH%;C:\Program Files (x86)\Rust\bin
+  - rustc -V
+  - cargo -V
+  - git submodule update --init --recursive
+
+build: false
+
+test_script:
+  - cargo build
+  - cargo test
```
mgattozzi: This should probably follow the same format as the travis script like testing npm etc.

steveklabnik: so funny fact https://ci.appveyor.com/project/steveklabnik/rustdoc/build/1.0.2#L142 due to https://github.com/steveklabnik/rustdoc/issues/22

so i can only build

I was going to test rustfmt and npm as well, but given that they should be platform-agnostic, figured it wasn't worth it. maybe they are...

mgattozzi: Hmmm. They should be. Personally though I don't like leaving things up to chance. Kind of funny that #22 is causing this to fail, but only on windows.

PR #49, opened 2017-07-19
Add error-chain to handle errors better

### LINKED ISSUE #46: Get rid of Box<Error>
We should have our own error type instead.

mgattozzi:
We were using `Box<std::error::Error>` for all of our error handling.
While not bad it makes dealing with them a bit of a pain and error-chain
handles errors very well, making it easy to create our own as well as
interface with other libraries errors. On top of that they end up
chaining well producing a stack of what caused what errors from the
point of origin all the way to the top level. This commit implements all
of the above to give us better more robust error handling.

@steveklabnik as this stand it's assuming the buildscript in #36 get's merged first but since it will be I built this code with that assumption. As you can see error-chain makes all the error handling easy to deal with and the results as well! Take a look and let me know what you think.

Closes #46

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,0 +1,14 @@
+error_chain! {
+    errors {
+        CrateErr(crate_name: &'static str) {
+            description("Crate not found")
+            display("Crate not found: \"{}\"", crate_name)
+        }
+    }
+
+    foreign_links {
+        Io(::std::io::Error);
+        Serde(::serde_json::Error);
+        Analysis(::analysis::AError);
+    }
+}
--- a/src/main.rs
+++ b/src/main.rs
@@ -5,6 +5,7 @@ use clap::{App, Arg, SubCommand};
 
 use rustdoc::{Config, build};
 
+use std::io::{Write, stderr};
 use std::process;
 use std::path::PathBuf;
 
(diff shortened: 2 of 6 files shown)


src/main.rs
```diff
@@ -56,8 +57,21 @@ fn main() {
...
+        let stderr = &mut stderr();
+        let errmsg = "Error writing to stderr";
+
+        writeln!(stderr, "Error: {}", e).expect(errmsg);
+
+        for e in e.iter().skip(1) {
+            writeln!(stderr, "Caused by: {}", e).expect(errmsg);
+        }
+
+        // The backtrace is not always generated. Try to run this example
```
mgattozzi: Should remove this and the other one saying example

src/error.rs
```diff
@@ -0,0 +1,14 @@
+error_chain! {
+    errors {
```
mgattozzi: These are our errors that we can create

src/error.rs
```diff
@@ -0,0 +1,14 @@
...
+        CrateErr(crate_name: &'static str) {
+            description("Crate not found")
+            display("Crate not found: \"{}\"", crate_name)
+        }
+    }
+
+    foreign_links {
```
mgattozzi: These are ones from a foreign source that implement `std::err::Error`

steveklabnik:
I like it; gonna sleep on it though!

Also, from merging your other PR, now has merge conflicts. 😄

mgattozzi:
Sounds good! Did a rebase and a push so this should be fine now! Take a look at it later when you get the chance! :D

mgattozzi:
@steveklabnik made some extra changes based off of the newest commits

steveklabnik:
The extra changes really help me see the value here, thank you!

mgattozzi:
You're welcome! :100: :D

PR #50, opened 2017-07-19
use cargo check instead of cargo build

steveklabnik:
Thanks @nrc https://internals.rust-lang.org/t/the-rustdoc-redux/5542/35


mgattozzi:
This should be fine. Only problem is appveyor failed, but it seems they're having SSL errors today.

steveklabnik:
https://twitter.com/RustStatus/status/887694846890307584

Issue #55, opened 2017-07-20
Inline docdata::build_data


steveklabnik:
In https://github.com/steveklabnik/rustdoc/pull/47, @jbcden raised some good points about `build_data` being awkward: https://github.com/steveklabnik/rustdoc/pull/47#discussion_r128135721

I've merged it, but this would still be a good refactoring.

https://github.com/steveklabnik/rustdoc/blob/master/src/lib.rs#L201-L242

jbcden:
@steveklabnik I'd be happy to take a run at this. I'm still fairly new to Rust so I might need help at some point.

steveklabnik:
@jbcden awesome! Please feel free to either open up a partial PR where we can discuss things or ping me on irc, whichever is more helpful to you, if you get stuck!

jbcden:
@steveklabnik was there more work around this that you wanted or is this issue closed by https://github.com/steveklabnik/rustdoc/pull/57?

steveklabnik:
It was! Thanks 😄

PR #57, opened 2017-07-21
Refactor build_data

jbcden:
This is a first pass at addressing https://github.com/steveklabnik/rustdoc/issues/55.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,6 +7,9 @@ extern crate error_chain;
 pub mod error;
 use error::*;
 
+pub mod item;
+use item::item::Metadata;
+
 use std::collections::HashMap;
 use std::fs::{self, File};
 use std::io;
@@ -168,7 +171,6 @@ fn generate_analysis(config: &Config) -> Result<()> {
 #[derive(Debug)]
 struct DocData {
     krate: Crate,
-    data: HashMap<String, Item>,
 }
 
 impl DocData {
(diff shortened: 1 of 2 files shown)


jbcden:
@steveklabnik Here's a first pass at refactoring the `build_data` function. I'm not sure if this is what you were thinking but let me know your thoughts!

src/lib.rs
```diff
@@ -321,10 +317,5 @@ struct Crate {
     id: analysis::Id,
     name: String,
     docs: String,
-    modules: Vec<String>,
-}
-
-#[derive(Debug)]
-enum Item {
-    Module { name: String, docs: String },
+    metadata: Vec<Metadata>,
```
steveklabnik: 👍

src/lib.rs
```diff
@@ -190,21 +192,15 @@ impl DocData {
...
     }
 
-    fn build_data(
-        config: &Config,
-        root_id: analysis::Id,
-        krate: &mut Crate,
-    ) -> Result<HashMap<String, Item>> {
-        let mut data = HashMap::new();
-
+    fn build_data(config: &Config, root_id: analysis::Id, krate: &mut Crate) -> Result<()> {
```
steveklabnik: So, this still ends up leaving this function and taking a `&mut Crate`; what do you think about possibly inlining this into the constructor? does it being its own function make sense still?

jbcden: @steveklabnik I go back and forth on that. As I said before, the main reason passing the crate in felt weird to me was that that seemed to signal that this function's main focus should be on the crate. Now that it is all about the crate, I don't feel this is too weird, though the name should definitely be changed to reflect this new role too.

steveklabnik: yeah let's just inline it for now 👍

src/lib.rs
```diff
@@ -213,14 +209,11 @@ impl DocData {
         for def in defs.iter() {
```
mgattozzi: I think it might make more sense to turn this from `iter` to `into_iter` that way the clones below can be avoided.

mgattozzi: The reason being is defs isn't used after this loop at all so consuming the values should be fine.

steveklabnik: yeah i was certainly cargo-culting this code; getting rid of eventual clones would be great!

jbcden:
> yeah let's just inline it for now 👍

@steveklabnik I inlined it in f91ddad.

jbcden:
@mgattozzi Thanks for the advice! I'm still pretty new to Rust so I'm sure I miss a lot of stuff like that. I switched to `into_iter` and removed the clones in f91ddad and d27a90e.

mgattozzi:
Woo awesome! :D

PR #60, opened 2017-07-22
Create missing asset sub-directories

### LINKED ISSUE #53: Code aborts when ember-fetch directory included in build script
Currently the build script [ignores](https://github.com/steveklabnik/rustdoc/blob/master/build.rs#L32) this directory but it would be nice to just have it work with ember-fetch and whatever other assets are there over time without this failure occurring. Something is causing it to fail during usage of `rustdoc` but I'm unsure as to what

hjr3:
The build process was failing when trying to include the `ember-fetch`
directory. The failure we related to the fact that we were trying to
create the `ember-fetch/fastboot-fetch-a.js` file without the
`ember-watch` director existing in `target/doc`.

Now `create_asset_file` assumes that `name` may be a path with
directories. It will first try to create the necessary directories
before trying to create the file.

Fixes #53

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6,7 +6,6 @@ dependencies = [
  "error-chain 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "indicatif 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "jsonapi 0.5.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


mgattozzi:
Ah awesome glad you found the problem

hjr3:
@mgattozzi I got your back homie!

steveklabnik:
Nice, thanks!

PR #62, opened 2017-07-23
Lookup crate name instead of package name

### LINKED ISSUE #51: Cannot find crate if crate name is different from package name
I have a crate where the package name is `feo-boy`, but the crate name is `feo_boy`. `rustdoc` fails to find the root crate in this case. Additionally, it returns the error message `Application error: Crate not found: "example"`, because of the hard-coded error.

hjr3:
The package name and crate name are often the same, but may be different.

Fixes #51

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,6 +1,6 @@
 error_chain! {
     errors {
-        CrateErr(crate_name: &'static str) {
+        CrateErr(crate_name: String) {
             description("Crate not found")
             display("Crate not found: \"{}\"", crate_name)
         }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -54,7 +54,7 @@ impl Config {
 pub fn build(config: &Config, artifacts: &[&str]) -> Result<()> {
     generate_analysis(config)?;
 
-    let package_name = package_name_from_manifest_path(&config.manifest_path)?;
+    let package_name = crate_name_from_manifest_path(&config.manifest_path)?;
     let data = DocData::new(&config.host, &package_name)?;
 
     let output_path = config.manifest_path.join("target/doc");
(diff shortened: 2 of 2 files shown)


hjr3:
discussing testing plan in https://github.com/steveklabnik/rustdoc/issues/51#issuecomment-317259613

steveklabnik:
I'm still drinking my coffee, but this is relying on the directory name to figure things out? That seems like it will mess up if that name gets changed. I thought we were gonna take the approach of reading the TOML file?

hjr3:
@steveklabnik my understanding of the issue is this:

- RLS uses crate name when doing the analysis.
- _crate name_ is defined as the name of the directory the crate is in. (Please correct me if I am wrong here. I cannot find _crate name_ explicitly defined in the crates.io documentation).
- The _name_ in the TOML file is the package name, which can be different from the crate name.

steveklabnik:
> crate name is defined as the name of the directory the crate is in. (Please correct me if I am wrong here. I cannot find crate name explicitly defined in the crates.io documentation).

In my understanding, crate name is the same as the package name unless you set it differently in your cargo.toml, that is, something like

```toml
[lib]
name = "foo"
path = "src/lib.rs"
```

hjr3:
@steveklabnik indeed that appears to be the case. ok let me try and fix this bug armed with the new information.

hjr3:
@euclio this should resolve the bug. Given the following `Cargo.toml` it will find the crate name `feo_boy`.

```toml
[package]
name = "feo-boy"

[dependencies]

[lib]
name = "feo_boy"
path = "src/lib.rs"
```

@steveklabnik this should resolve the bug, though my PR is failing on a frontend issue.

mgattozzi:
If you rebase off master and do a force push it should resolve the failing test since this was fixed earlier.

hjr3:
@mgattozzi strange, i was. i did re-push and travis passed. 🤷‍♂️

mgattozzi:
@hjr3 that's just weird. Wonder if it was a fluke or something

steveklabnik:
Thank you!

PR #63, opened 2017-07-24
Revert appveyor fix since upstream fixed it

mgattozzi:
appveyor builds were failing due to ssl issues but they've been fixed
upstream. We don't need this workaround anymore:

https://github.com/appveyor/ci/issues/1669


steveklabnik:
Thank you!

PR #66, opened 2017-07-25
Improve fetching the defs

steveklabnik:
EDIT: the below was true when i opened this PR, with rayon, but I've since pulled it out.

---------------------

thank you based @sp3d for helping me get this to compile

The next big step is to start fetching more than just the first-level defs. 😄 

I'm opening this PR, not because I plan on merging, but because I think it's not great and wondered if others had good feedback. I talked to @nikomatsakis about this earlier, and there's two problems here:

1. it's not actually clear that doing this with rayon is going to be much faster, as there's not that much work to actually do
2. this is sort of half-way between a "pure channels" model and a "pure rayon" model, so it kind of inherits the worst of both worlds.

I'm going to continue working on the next steps after this (specifically, while this now renders all of the submodules, they're not nested, because the JSON generating code is also bad) but in the meantime, if any of you want to hack on improving this code, feel free to leave comments or send a PR to my PR 😄

### DIFF
--- a/example/src/lib.rs
+++ b/example/src/lib.rs
@@ -33,3 +33,14 @@ pub fn sample_function(x: i32) -> ExampleStruct {
 pub mod example_module {
     pub struct AnotherStruct;
 }
+
+/// nested 1
+pub mod nested1 {
+    /// nested 2
+    pub mod nested2 {
+        /// nested 3
+        pub mod nested3 {
+
+        }
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -206,7 +206,25 @@ impl DocData {
             metadata: Vec::new(),
         };
 
-        let defs = host.for_each_child_def(root_id, |_, def| def.clone())?;
+        // TODO: https://github.com/steveklabnik/rustdoc/issues/70
+        fn recur(id: &analysis::Id, host: &AnalysisHost) -> Vec<analysis::Def> {
+            let defs_and_ids = host.for_each_child_def(*id, |id, def| (id, def.clone()))
+                .unwrap();
+
+            let mut v = Vec::new();
+
+            for (id, def) in defs_and_ids.into_iter() {
+                v.push(def);
+
+                for def in recur(&id, host).into_iter() {
+                    v.push(def);
+                }
+            }
+
+            v
+        }
+
+        let defs: Vec<analysis::Def> = recur(&root_id, host);
 
         for def in defs.into_iter() {
             match def.kind {


mgattozzi:
To address 1 Rayon works on the possibility of there being work but it doesn't always split it up into parallel depending on the size where iterative would be faster. If there's a larger dataset I'm sure the benefits would be seen more.

What other parts are you trying to make parallel with Rayon? Also your tests failed because of rustfmt :(

steveklabnik:
> If there's a larger dataset I'm sure the benefits would be seen more.

Yeah I'm thinking like, documenting servo, or the stdlib.

> What other parts are you trying to make parallel with Rayon?

Eventually, the JSON generation too; i'm gonna have to throw out the JSON API package I'm using, I think.

mgattozzi:
Is it too slow? Would we have to roll our own?

steveklabnik:
> Is it too slow? Would we have to roll our own?

It's not the speed, though it's not as fast as it could be (it uses a lot of HashMaps and such).

There are two issues:

1. you have to build the entire tree in memory, which is awkward for super-large crates (again, thinking of servo and the stdlib here)
2. it supports all of the various features of the spec, which we don't actually need. we pretty much need relationship links and that's it. Because of this, I don't think it's too hard to implement the tiny chunk with serde-json straight up, rather than relying on the whole library.

basically, https://github.com/steveklabnik/rustdoc/blob/master/src/lib.rs#L242-L313 feels like a *lot* of boilerplate for something that should be much easier.

mgattozzi:
I think we could easily make what we need then. If you outline the format for the JSON in the issue tracker whenever you have the chance I could look into replacing it.

steveklabnik:
@mgattozzi that'd be awesome except that it is sorta what I'm working on right now, so stuff might clash a bit.  That said, we could divide and conquer this!

The next big steps are:

1. taking this list of all defs and properly generating the relationships between them
2. from there, turning that into JSON
3. modifying the front-end routes to be able to handle nested modules

If you want to tackle 2, that'd be great as I can work on 1 and/or 3 while you hack on it. Or, if either of those others sound good, I can work on the ones you don't want to 😄

mgattozzi:
I could do 2. This would be using serde by itself and not the crate right? What would that need to look like exactly?

steveklabnik:
Done in https://github.com/steveklabnik/rustdoc/issues/69

steveklabnik:
Merging with the non-rayon version, filed https://github.com/steveklabnik/rustdoc/issues/70 for further improvements.

PR #68, opened 2017-07-25
Test fixups

steveklabnik:
In https://github.com/steveklabnik/rustdoc/pull/61, @euclio added a test runner. In https://github.com/steveklabnik/rustdoc/pull/67 i got them building on Windows.

Adding a second test was failing, because keeping the same TempDir for every test meant that old files were lying around, which made the runner confused. The easiest thing is to make a temporary directory per test.

I also added a second test, testing module stuff. I'm going to build on this for the nested module stuff I'm working on next.

@euclio what do you think?

### DIFF
--- a/tests/source.rs
+++ b/tests/source.rs
@@ -208,9 +208,9 @@ fn run_test(json: &serde_json::Value, pointer: &str, regex: &Regex) -> Result<()
 #[test]
 fn source() {
     let source_dir = Path::new("tests/source");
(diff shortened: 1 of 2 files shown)


steveklabnik:
(I'm merging this for now, but @euclio I'd still like to hear what you think about this change 😄 )

euclio:
Looks good to me. I'm working on a patch to generate tests as separate `#[test]` functions, so this would've needed to be done anyways!

Issue #69, opened 2017-07-25
Replace jsonapi crate with plain serde_json


steveklabnik:
I started off using the jsonapi crate to get going, but given that we don't use all of the spec, it's a little complex for our needs. In addition, it uses a lot of dynamic dispatch, hashmaps, etc that we wouldn't need.

Fundamentally, the JSON looks like this:

```json
{
  "data": {
    "type": "crate",
    "id": "example",
    "attributes": {
      "docs": " docs go here"
    },
    "relationships": {
      "modules": {
        "data": [
          {
            "type": "module",
            "id": "example::example_module"
          }
        ]
      }
    }
  },
  "included": [
    {
      "type": "module",
      "id": "example::example_module",
      "attributes": {
        "docs": " docs go here",
        "name": "example_module"
      }
    }
  ]
}
```

Let's break this down a bit:

at the top level, we have two keys: `data` and `included`. `included` is an array of documents, `data` is a single one. `data` is for the crate itself, `included` is for everything else.

Each document looks like this:

```json
{
    "type": "crate",
    "id": "example",
    "attributes": {
      "docs": " docs go here"
    },
    "relationships": {
      "modules": {
        "data": [
          {
            "type": "module",
            "id": "example::example_module"
          }
        ]
      }
    }
}
```

There are four keys:

1. "type" contains the type, so right now, the top level is "crate" and everything else is "module". Eventually this will expand to the other items, but we don't do that just yet.
2. "id" is called the "qualified name" in rls-analysis. A globally unique identifier. More about this in "relationships" below.
3. "attributes" contains all of the actual data for the thing. Right now, we only have two attributes: "docs", which is the doc string, and "name", which is just the name, without the parent stuff. compare the name to the id of the included module above, for example.
4. "relationships" establishes the links between stuff. In the first example, there are no sub-crates, and so it has no `relationships`, the only one is in the crate itself. If `example_module` had a sub-module, it'd have a relationship to it; basically, the stuff without relationships is the leafs of the tree. You can also think of this as a foreign key relationship in a database.

Some more on relationships, since they have their own structure:

```json
      "modules": {
        "data": [
          {
            "type": "module",
            "id": "example::example_module"
          }
        ]
      }
```

At the top levels are keys for the types of relationships that they have; in this case, we only have modules, so there's only one, but that expands in the future.

[...]

PR #73, opened 2017-07-26
Improvement to frontend routes

steveklabnik:
This is the start of the work for the frontend to be able to do nesting.

Much thanks to @locks and @fivetanley for some help getting this going; it *almost* works, but not quite.

That is, it works, but the URLs still have some `::`s in them as opposed to `/`s. Still gotta sort out where exactly things aren't 100% there yet.

This also expands the fixture to contain the proper nested data as described in #69 ; and it's now pretty-printed. I'll probably un-pretty-print it before I end up merging.


frontend/app/routes/modules.js
```diff
@@ -1,7 +1,22 @@
 import Ember from 'ember';
 
+/*
 export default Ember.Route.extend({
     model({ crate_name, path }) {
         return this.get('store').peekRecord('module', `${crate_name}::${path}`);
     }
 });
+*/
```
steveklabnik: all of this commented-out stuff will need to go

frontend/app/routes/modules.js
```diff
@@ -1,7 +1,22 @@
...
+
+export default Ember.Route.extend({
+  serialize(mod) {
+    // convert :: to /
+    return mod.get('id').replace(/::/g, '/');
+  },
+  model(params) {
+    // convert from / to :: //
+    const moduleId = `${params.crate_name}::${params.path.replace(/\//g, '::')}`;
+//    return this.store.findRecord('module', moduleId, {reload: false});
```
steveklabnik: this still needs to be removed

frontend/app/templates/modules.hbs
```diff
@@ -1,4 +1,17 @@
 <h1>Module {{ model.name }}</h1>
 {{markdown-render model.docs }}
 
+<h2>Submodules</h2>
+
+<ul>
+{{#each model.child_modules as |mod|}}
+  <li>
+    {{#link-to 'modules' model.id mod.name }}
```
steveklabnik: this is still off somehow

steveklabnik:
This is now fixed up! The big thing was to realize that a module already has the crate name in front, so by dropping that, it all just works 🎊

Issue #75, opened 2017-07-27
Write more tests


steveklabnik:
Now that we have a test framework, we should verify everything in the JSON. Right now, we only have two tests: https://github.com/steveklabnik/rustdoc/tree/master/tests/source

testing all aspects of https://github.com/steveklabnik/rustdoc/blob/master/frontend/public/data.json is good; that'd take two steps:

1. we should use `jsonapi` to validate that it's proper JSON-API
2. we should test that the proper relationships are created

DevOrc:
For the first one were thinking of something like the [validate function](https://docs.rs/jsonapi/0.5.2/jsonapi/api/struct.JsonApiDocument.html#method.validate)? If so, I  can probably write a quick PR for this!

steveklabnik:
@DevOrc yup! And also checking the error of [`from_str`](https://docs.rs/jsonapi/0.5.2/jsonapi/api/struct.JsonApiDocument.html#method.from_str), since that's how you'd load it up.

mgattozzi:
1 has been completed. For those wishing to help add tests only 2 needs to be done to help close out this issue! :D

steveklabnik:
Marking this one as easy because it fundamentally comes down to writing some `//` lines, like these ones: https://github.com/steveklabnik/rustdoc/blob/master/tests/source/modules.rs#L3

If how to do this isn't clear I'm quite happy to work with anyone on this!

sedrik:
I started looking at writing this test but I have some questions.

```rust
#![crate_type = "lib"]

//! An example crate.
//!
//! This crate is used to test out the new rustdoc.
//!
//! It should have an example of almost everything, for testing
//! purposes.

// @has /data/type 'crate'
// @has /data/id 'nested_modules'

// @has /data/attributes/docs ' An example crate.\n\n This crate is used to test out the new rustdoc.\n\n It should have an example of almost everything, for testing\n purposes.\n'
```

This fails as the json that gets parsed is

```
{
  "data": {
    "attributes": {
      "docs": " An example crate."
    },
    "id": "nested_modules",
    "type": "crate"
  },
  "included": []
}
```

Did I write the documentation string wrong? also should the leading space be included for each line? and why does the string at `/data/attributes/docs` end with a \n (it would make sense to skip that right?).

steveklabnik:
@sedrik that seems... odd.

>  (it would make sense to skip that right?).

We don't do any processing of the docs string on the backend, leaving that entirely to the frontend.

[...]

PR #83, opened 2017-07-28
Added a test to validate json

DevOrc:
Fixes part one of #75.

 Question: Should this test be placed in a different module? For example have a source_tests module and generic_tests module

### DIFF
--- a/tests/source.rs
+++ b/tests/source.rs
@@ -33,13 +33,13 @@ extern crate jsonapi;
 use std::fs::{self, File};
 use std::io::{self, BufReader};
 use std::io::prelude::*;
-use std::path::{Path};
+use std::path::Path;
 use std::process::Command;
 
 use analysis::{AnalysisHost, Target};
 use regex::Regex;
 use serde_json::Value;
-use rustdoc::{DocData};
+use rustdoc::DocData;
 
 error_chain! {
     foreign_links {
--- a/tests/source.rs
+++ b/tests/source.rs
@@ -28,18 +28,20 @@ extern crate serde_json;
 extern crate regex;
 extern crate rls_analysis as analysis;
 extern crate tempdir;
+extern crate jsonapi;
 
 use std::fs::{self, File};
 use std::io::{self, BufReader};
 use std::io::prelude::*;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 
 use analysis::{AnalysisHost, Target};
 use regex::Regex;
 use serde_json::Value;
+use jsonapi::api::{JsonApiDocument};
 
-use rustdoc::DocData;
+use rustdoc::{DocData, build, Config};
 
 error_chain! {
     foreign_links {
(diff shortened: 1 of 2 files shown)


steveklabnik:
Thank you! This looks great.

CI failed due to rustfmt issues.

> Question: Should this test be placed in a different module? For example have a source_tests module and generic_tests module

Yeah, let's put this in a `tests/validation.rs`.

tests/source.rs
```diff
@@ -314,4 +316,42 @@ mod tests {
...
+
+        let result = build(&config, &["json"]);
+
+        if let Err(ref e) = result {
+            panic!("Error: {}", e);
+        }
+
+        let path = Path::new("example/target/doc/data.json");
+
+        let mut file = match File::open(&path) {
```
steveklabnik: this match could be `unwrap_or_else` like above

tests/source.rs
```diff
@@ -314,4 +316,42 @@ mod tests {
...
+            Err(e) => panic!("couldn't open data.json: {}", e),
+            Ok(file) => file,
+        };
+
+        let mut s = String::new();
+        let result = file.read_to_string(&mut s);
+
+        if let Err(ref e) = result {
```
steveklabnik: this could also be `unwrap_or_else` on the `read_to_string` call

tests/source.rs
```diff
@@ -314,4 +316,42 @@ mod tests {
...
+            panic!("Error: {}", e);
+        }
+
+        let result = JsonApiDocument::from_str(&s);
+        let doc = match result{
```
steveklabnik: same here

tests/source.rs
```diff
@@ -314,4 +316,42 @@ mod tests {
...
+            Err(e) => panic!("Error: {}", e),
+            Ok(doc) => doc
+        };
+
+        match doc.validate() {
```
steveklabnik: and here 😄

DevOrc: How would unwrap_or_else work here? It should return none if no errors are found.

tests/validation.rs
```diff
@@ -0,0 +1,41 @@
...
+        let mut s = String::new();
+        file.read_to_string(&mut s).unwrap_or_else(|err| {
+            panic!("Error: {}", err);
+        });
+
+        let doc = JsonApiDocument::from_str(&s).unwrap_or_else(|err| {
+            panic!("Error: {}", err);
+        });
+
+        match doc.validate() {
```
DevOrc: How would unwrap_or_else work here? It should return none if no errors are found.

steveklabnik: oops, you're right!

steveklabnik:
@DevOrc one other thing; it looks like you may not have set up your git properly to link to github 😄

tests/validation.rs
DevOrc: I thinks its right now!

DevOrc:
I thinks its right now! Do I have to recommit it to update the authorship?

mgattozzi:
This a small nitpick about the title, but using `Add` instead of `Added` makes for a [better commit message](https://chris.beams.io/posts/git-commit/#imperative). Otherwise this is some good code @DevOrc!

mgattozzi:
@steveklabnik not sure if this will conflict with #92 now or not.

steveklabnik:
Thank you!

@mgattozzi we can still use it as a dev-dependency; i'll comment over on that PR

PR #88, opened 2017-07-29
Fix #87: remove nested item module.

### LINKED ISSUE #87: get rid of nested item module
We have https://github.com/steveklabnik/rustdoc/blob/master/src/item.rs#L1  which leads to https://github.com/steveklabnik/rustdoc/blob/master/src/lib.rs#L14

that first line should be removed, and the import adjusted, so it's just `item` and not `item::item`.

This is a very good first issue!

jasondavies:


mgattozzi:
This looks good to me @steveklabnik. It does everything described in #87.

steveklabnik:
Thanks so much @jasondavies !

PR #91, opened 2017-07-30
Document all currently existing items in rustdoc

mgattozzi:
Documentation is important, especially if onboarding new contributors.
`rustdoc` had little documentation on it's types and functions. This
commit does two things. First it adds documentation for everything being
documented and adds a `#![deny(missing_docs)]` lint (the exception being
`error.rs` but some documentation was left there). The first part gives
us documentation on everything in the crate today. The second part makes
sure that we continuously document new things as they are added to the
project. This makes it easier to handle documenting things in the long
run by just adding them sooner rather than later.

@steveklabnik I'd like to wait till #88 gets merged so I can get rid of the temp line around the module before this gets merged. I think I covered everything here well but I'm sure we can figure out a better way to say things then how I wrote it if something needs tweaking. Also let me know how you feel about `#![deny(missing_docs)] but I think this would be good to have on. Especially for a documentation tool


DevOrc:
Thank you! When I was originally looking through the repository, I was annoyed by the lack of comments!

steveklabnik:
> I'd like to wait till #88 gets merged 

It's now done! And some other PRs too, so there's a lot of conflicts 😈 

> Also let me know how you feel about `#![deny(missing_docs)] but I think this would be good to have on. Especially for a documentation tool

So, for my own projects I leave it off, since I've certainly done `/// lol` before to get things passing. This happens because during development, having a ton of warnings is annoying.

What if we did something slightly different: we leave it off, but in travis, instead of `cargo build`, we run `cargo rustc --deny missing_docs` (or whatever the flag is), so that we don't have to deal with it while developing, but Travis will catch it?

mgattozzi:
Yeah I think that would be good that way people can develop without writing bad docs but have this be a lint when building.

mgattozzi:
I'll probably work on this more once we get #92 in since there's a lot of stuff there I need to document but didn't and it probably needs a little explaining about why I did what.

euclio:
> >Also let me know how you feel about `#![deny(missing_docs)] but I think this would be good to have on. Especially for a documentation tool

> So, for my own projects I leave it off, since I've certainly done /// lol before to get things passing. This happens because during development, having a ton of warnings is annoying.

I think `#![warn(missing_docs)]` is a good compromise. Rust tends to have a good culture of removing warnings, and then it doesn't get too much in the way during development.

mgattozzi:
@euclio I'd be good with that!  We could then still deny on the PR as well. @steveklabnik thoughts?

steveklabnik:
>  it doesn't get too much in the way during development.

I find this kind of thing pretty annoying, but am willing to believe I'm in the minority, and it's always easy for me to turn them off temporarily too. Seems good!

steveklabnik:
Other PR has now merged; this could use a rebase!

mgattozzi:
Yup a rebase some additions and other changes based off the comments here!

mgattozzi:
So adding the deny with `cargo rustdoc` was ignored since there was the `warning` attribute. I think that should suffice for now.

steveklabnik:
Hm, what about all of this output in the build log: https://travis-ci.org/steveklabnik/rustdoc/builds/259458993#L623

src/json.rs
```diff
@@ -5,20 +5,69 @@ use std::collections::HashMap;
...
+/// Number of possible values in the enum item::Metadata
 pub const METADATA_SIZE: usize = 14;
 
-// Under each item in data.relationships how many fields are there
-// For now we only have one which is data.
+/// Under each item in data.relationships how many fields are there. For now we only have one which
+/// is data. As we might add more this number should increase
 const DATA_SIZE: usize = 1;
 
 #[derive(Serialize, Debug)]
```
steveklabnik: i'd prefer to put this after the doc comments, not before

src/json.rs
```diff
@@ -5,20 +5,69 @@ use std::collections::HashMap;
...
+/// The data structure used to serialize crate documentation data to JSON for consumption by the
+/// front end. When fully serialized the data looks something like:
+///
+/// ```ignore, json
```
steveklabnik: you shouldn't need `ignore` here, just `json`.

src/json.rs
```diff
@@ -5,20 +5,69 @@ use std::collections::HashMap;
...
+///                 "docs": "docs go here",
+///                 "name": "example_module"
+///             }
+///         }
+///     ]
+///}
+///```
+///
+/// ## Fields
+/// - `data`: The top level crate information and documentation
```
steveklabnik: put an extra `///` above this line to let the heading breathe, please!

src/json.rs
```diff
@@ -5,20 +5,69 @@ use std::collections::HashMap;
...
+/// - `included`: The rest of the crate information and documentation
+///
```
steveklabnik: ✂️

src/json.rs
```diff
@@ -5,20 +5,69 @@ use std::collections::HashMap;
...
 pub struct Documentation<'a> {
     data: Option<Document<'a>>,
     included: Option<Vec<Document<'a>>>,
 }
 
 #[derive(Serialize, Debug)]
```
steveklabnik: same here, please put this below

src/json.rs
```diff
@@ -5,20 +5,69 @@ use std::collections::HashMap;
...
+/// A sub type of the `Documentation` struct. It contains the majority of the data. It can be used
+/// for both the `data` and `included` field in the serialized JSON.
+///
+/// ## Fields
+/// - `ty`: The type of the the item (e.g. "crate", "function", "enum", etc.)
+/// - `id`: The unique identifier associated with this item
+/// - `attributes`: The attributes associated with the item like documentation or it's name
+/// - `relationships`: An optional field used to show the relationship between the crate to the
+///   otheritems in the crate
+///
```
steveklabnik: ✂️

src/json.rs
```diff
@@ -5,20 +5,69 @@ use std::collections::HashMap;
...
     data: Option<Document<'a>>,
     included: Option<Vec<Document<'a>>>,
 }
 
 #[derive(Serialize, Debug)]
+/// A sub type of the `Documentation` struct. It contains the majority of the data. It can be used
+/// for both the `data` and `included` field in the serialized JSON.
+///
+/// ## Fields
+/// - `ty`: The type of the the item (e.g. "crate", "function", "enum", etc.)
```
steveklabnik: `///` above here too

src/json.rs
```diff
@@ -28,32 +77,44 @@ pub struct Document<'a> {
 }
 
 #[derive(Serialize, Debug)]
```
steveklabnik: same three things here, i'll stop saying it 😄

src/lib.rs
```diff
@@ -1,3 +1,7 @@
+#![warn(missing_docs)]
+
+//! Functions used to generate the documentation for Rust Crates.
```
steveklabnik: i'd move this above the warning

mgattozzi:
@steveklabnik Not sure why that's even showing up. My guess is it's an error-chain bug causing it since it seems to also affect the identicaf crate but I'll need to check. As for the formatting it shouldn't be too hard to clean up!

PR #92, opened 2017-07-31
Remove jsonapi and replace it with serde

### LINKED ISSUE #69: Replace jsonapi crate with plain serde_json

mgattozzi:
This changes rustdoc over completely to serde. While jsonapi was a great
crate to get started it had many features and things we weren't going to
use at all. Instead we implemented a limited subset of the spec. This
moves us over to just using serde to output data about the crate. A few
things worth noting:

- The `Crate` type was merged into the `DocData` type since it was only
  used in that type. This made the resulting serialization code much
  cleaner by removing references to krate.
- The builder method was used as a way to cleanly build up the types
  for eventual serialization and because it's readable.
- Using `&str` vs String was chosen to avoid allocations that weren't
  needed
- `HashMaps` were made with the capacity of given constants since we
  know how many items at most will exist there.

Overall this commit makes serializing data easier to work with and
more reabable then before since we don't need to use all of jsonapi's
more advanced features. The nice thing is we can still use jsonapi to
validate the output that we create!

Closes #69 (Nice)

@steveklabnik Unfortunately this is a bit bigger of a PR and harder to review due to the nature of having to rip things out and replacing them. Hopefully it's not too bad but please leave any feedback and I'll be happy to update this PR.


src/json.rs
```diff
@@ -0,0 +1,115 @@
...
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Document {
+    #[serde(rename = "type")]
+    type_: String,
+    id: String,
+    attributes: HashMap<String, String>,
+    relationships: Option<HashMap<String, HashMap<String, Vec<Data>>>>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
```
mgattozzi: I can actually get rid of `Clone` here. It's not needed anymore.

DevOrc:
This will probably conflict with #83 because it uses the jsonapi to validate it... So you could probably just add it as a dev-dependency? I don't know if there is a equivalent of the [validate function](https://docs.rs/jsonapi/0.5.2/jsonapi/api/struct.JsonApiDocument.html#method.validate) is serde...

mgattozzi:
I put it in as a dev dependency and it worked perfectly fine as is locally :D No need to worry about it now.

steveklabnik:
Yup!

@mgattozzi I'm going to review this shortly, it's the biggest PR of the weekend ones so it was easier to knock those out!

mgattozzi:
@steveklabnik Not a problem! I just rebased and cleaned everything up and pushed it with all the changes I made rolled into the one commit for you to review.

@DevOrc thanks for the json validation test. Definitely helps knowing that what I cooked up actually works correctly :D

src/json.rs
```diff
@@ -0,0 +1,118 @@
+//! Code used to serialize crate data to JSON. We use a subset of the JSON-API spec.
+//!
```
steveklabnik: ✂️

src/json.rs
```diff
@@ -0,0 +1,118 @@
...
+#[derive(Serialize, Debug)]
+pub struct Documentation<'a> {
+    data: Option<Document<'a>>,
+    included: Option<Vec<Document<'a>>>,
+}
+
+#[derive(Serialize, Debug)]
+pub struct Document<'a> {
+    #[serde(rename = "type")]
+    type_: &'a str,
```
steveklabnik: i'd prefer `ty`

steveklabnik:
:tada: thanks so much!

mgattozzi:
You're welcome :D

PR #94, opened 2017-07-31
Parallelize the recur function

### LINKED ISSUE #70: Improve fetching of defs
In https://github.com/steveklabnik/rustdoc/pull/66 we did a naive recursive fetching of all defs, but for large crates, this might blow the stack. In addition, it's all serial; we may want to investigate parallelizing it.

mgattozzi:
Prior to this recur kept being called. This might have led to a blown
stack on larger projects (still unsure). However, it wasn't all too fast
to run and could easily have been parallelized. This commit does just
that by grabbing all of the analysis in parallel for everything
possible. It then reduces them into a single Vec and then extends are
already existing top level Vec into one just like before.

This was blazingly fast in release mode but debug seemed to be a bit
slow (most likely because iterators usually are in debug).

Closes #70 

@steveklabnik I'd be open to further improving this and continued testing

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,6 +8,7 @@ dependencies = [
  "jsonapi 0.5.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "quote 0.3.15 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rayon 0.8.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "rls-analysis 0.4.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.11 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


mgattozzi:
@steveklabnik looks like appveyor is acting wonky again today :(

mgattozzi:
I did some testing vs the master branch on the example crate:
```
        master        rayon
debug   3.80s         3.825s
release 0.65s         0.64s
```

As you can see the results are hardly any different. I think we would see bigger changes with a larger data set or crate. Rayon's philosiphy is that it *might* run in parallel if it would be faster and default to serial iteration otherwise when that's faster. For smaller datasets it tends to not run in parallel (from what I've witnessed in my own crates). It might be good to expand the example crate or run this on something like Servo and see what we get, but for now the tool is hardcoded to example

steveklabnik:
Yeah, I don't expect much difference here for a small crate. That said, I do want to do something similar with the JSON generation, which I expect to be more of a big deal.

I thought we had already fixed the hardcoding?

mgattozzi:
Ah we did. I just thought we hadn't that's my bad. I could do another PR after this one to parallelize it somehow, though I think that might be a little harder to do.

steveklabnik:
>  I could do another PR after this one to parallelize it somehow, though I think that might be a little harder to do.

Naw I'm about to *significantly* revise this code.

Let's see if I can sort AppVeyor and then we can merge.

mgattozzi:
@steveklabnik Oh I look forward to that code showing up in the PR queue.

PR #104, opened 2017-08-03
Let rustdoc work with workspaced crates [WIP]

mgattozzi:
rustdoc was unable to work with workspaced crates before. This commit
now allows that by searching for all workspaced crates and generating
docs for them one at a time.

This is a complete work in progress but I'd like to solicit feedback as I work on it. Namely I've run into a few problems so far:

- Fails on non lib dirs
- Fails for other reasons that I'm still debugging. (Try this against the rayon crate to see what I mean)

Either way I do want to parallelize this as well after the kinks are worked out. Was wondering if the output should be closer to what cargo doc currently outputs or do we want to keep what we have?

### DIFF
--- a/src/cargo.rs
+++ b/src/cargo.rs
@@ -30,12 +30,25 @@ pub fn generate_analysis(manifest_path: &Path) -> Result<()> {
     let output = command.output()?;
 
     if !output.status.success() {
-        return Err(
-            ErrorKind::Cargo(
-                output.status,
-                String::from_utf8_lossy(&output.stderr).into_owned(),
-            ).into(),
-        );
+        let mut command = Command::new("cargo");
+        command
+            .arg("check")
+            .arg("--bins")
+            .arg("--manifest-path")
+            .arg(manifest_path.join("Cargo.toml"))
+            .env("RUSTFLAGS", "-Z save-analysis")
+            .env("CARGO_TARGET_DIR", manifest_path.join("target/rls"));
+
+        let output = command.output()?;
+
+        if !output.status.success() {
+            return Err(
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 target/
 **/*.rs.bk
+src/asset.in
```
steveklabnik: have you pulled in recent changes? this file is now in `target`, so we don't need this line

src/cargo.rs
```diff
@@ -113,6 +113,68 @@ fn crate_name_from_metadata(metadata: &serde_json::Value) -> Result<String> {
     )
 }
 
+
+/// Grab the name of the binary or library from it's `Cargo.toml` file.
```
steveklabnik: its

mgattozzi:
I've made some more changes so I'll push them later.

.gitignore
mgattozzi: I have the latest changes. Guess I just never deleted the file from before

mgattozzi:
I thought being able to document binary crates in workspaces would be good but it looks like cargo doesn't do that right now. I'll just need to figure out a way to skip them without throwing an error.

euclio:
You might be running into #105 while trying to generate docs for binaries.

mgattozzi:
@euclio Possibly might be the case. I run it with the `bins` flag but I keep getting a crate not found error for the binary one despite it's def existing in another one when generating the JSON. Unsure as to why

mgattozzi:
I don't want to try to rebase this right now so I'll just rewrite it all later

PR #106, opened 2017-08-03
Remove DocData & Item

steveklabnik:
Instead, build the Documentation up itself. This means that we now have
to have Documentation own its data, which feels like adding more
allocations, but the reason we need to do this is that item itself was
allocating.


src/lib.rs
```diff
@@ -150,129 +147,75 @@ fn generate_and_load_analysis(config: &Config) -> Result<()> {
...
-            let item_relationships = relationships.entry(relations_key).or_insert_with(
-                Default::default,
-            );
-            item_relationships.push(Data::new().ty(ty).id(&item.qualified_name));
-        }
-
-        let len = self.name.len();
+    let mut data_document = Document::new()
+        .ty(String::from("crate"))
+        .id(crate_name.to_string())
```
mgattozzi: Why is this `to_string()` instead of `String::from()`? On a side note why not just use `into()`? For more explicitness?

steveklabnik: so, standard style is `to_string` for everything. I still prefer `String::from` for literals.

Yeah, into is a bit too magic for me

mgattozzi: I can see that. `to_owned()` is also a good one I think, but I'm fine with a more standarized style in the codebase.

PR #108, opened 2017-08-03
Properly nest definitions

steveklabnik:
This is still a work in progress, right now it's only a failing test. I took a short stab at this just now, but I don't think I'm gonna get it done today.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,7 +93,6 @@ pub fn build(config: &Config, artifacts: &[&str]) -> Result<()> {
         let mut file = File::create(json_path)?;
         file.write_all(json.as_bytes())?;
         spinner.finish_with_message("Generating JSON: Done");
-        println!("{}", json);
     }
 
     // now that we've written out the data, we can write out the rest of it
@@ -162,8 +161,7 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
         let def = host.get_def(*id)?;
 
         let mut ids = Vec::new();
-        host.for_each_child_def(*id, |id, _| { ids.push(id); })
-            .unwrap();
+        host.for_each_child_def(*id, |id, _| ids.push(id))?;
 
         let ty = match def.kind {
             DefKind::Mod => String::from("module"),
@@ -194,7 +192,7 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
 
         // create child relationships
         let relationships: Vec<_> = child_docs
-            .iter()
+            .par_iter()
             .filter(|doc| doc.ty == "module")
             .map(|doc| Data::new().ty(doc.ty.clone()).id(doc.id.clone()))
             .collect();
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -162,9 +162,8 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
         let def = host.get_def(*id)?;
 
         let mut ids = Vec::new();
-        host.for_each_child_def(*id, |id, _| {
-            ids.push(id);
-        }).unwrap();
+        host.for_each_child_def(*id, |id, _| { ids.push(id); })
+            .unwrap();
 
         let ty = match def.kind {
             DefKind::Mod => String::from("module"),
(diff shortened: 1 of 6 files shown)


steveklabnik:
zomg this is now passing tests.

I think there's still a bug, in that I think that it will add _too many_ relationships, but pushing up what I have now for feedback.

src/lib.rs
```diff
@@ -94,6 +93,7 @@ pub fn build(config: &Config, artifacts: &[&str]) -> Result<()> {
         let mut file = File::create(json_path)?;
         file.write_all(json.as_bytes())?;
         spinner.finish_with_message("Generating JSON: Done");
+        println!("{}", json);
```
steveklabnik: this is erroneus

src/lib.rs
```diff
@@ -158,62 +158,84 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
 
     let root_def = host.get_def(root_id)?;
 
-    fn recur(id: &analysis::Id, host: &AnalysisHost) -> Vec<analysis::Def> {
+    fn recur(id: &analysis::Id, host: &AnalysisHost) -> Result<Vec<Document>> {
+        let def = host.get_def(*id)?;
+
         let mut ids = Vec::new();
-        let mut defs = host.for_each_child_def(*id, |id, def| {
+        host.for_each_child_def(*id, |id, _| {
```
steveklabnik: maybe this is a mis-use of the API now, dunno

src/lib.rs
```diff
@@ -162,9 +162,8 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
         let def = host.get_def(*id)?;
 
         let mut ids = Vec::new();
-        host.for_each_child_def(*id, |id, _| {
-            ids.push(id);
-        }).unwrap();
+        host.for_each_child_def(*id, |id, _| { ids.push(id); })
+            .unwrap();
```
steveklabnik: this should be ?

src/lib.rs
```diff
@@ -158,62 +157,82 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
...
-        );
-        item_relationships.push(Data::new().ty(ty).id(def.qualname));
+        let mut docs = Vec::new();
+
+        let child_docs: Vec<Document> = ids.into_par_iter()
+            .map(|id: analysis::Id| recur(&id, host))
+            .reduce(|| Ok(Vec::new()), |a: Result<Vec<Document>>,
+             b: Result<Vec<Document>>| {
+                let mut a = a?;
+                let b = b?;
```
mgattozzi: you could rewrite this by declaring `|mut a: Result<` for the closure and  then rewrite the body as:

```
a?.extend(b?);
a
```

I think

mgattozzi: or you might still need to do the unwrapping still with the `let mut a` stuff

src/lib.rs
```diff
@@ -158,62 +157,82 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
...
+
+                a.extend(b);
+                Ok(a)
+            })?;
+
+        // create child relationships
+        let relationships: Vec<_> = child_docs
+            .iter()
```
mgattozzi: you can do par_iter() here as well.

src/lib.rs
steveklabnik: error[E0382]: use of moved value: `a`
   --> src\lib.rs:187:17
    |
186 |                 a?.extend(b?);
    |                 - value moved here
187 |                 a
    |                 ^ value used here after move

mgattozzi:
🎆 🎈 This looks good to me so far. 🎈 🎆

steveklabnik:
So, in the last commit, you can see one of the things that's not right yet: it renders this:

modules

That is, it only shows the last chain, rather than both. If you remove the next stuff, it only renders the nested stuff, not `example_module`.

steveklabnik:
To be even clearer, they exist in the `included` array properly, it's the root crate's association that's not proper.

mgattozzi:
Ah so that was the bug you were talking about. Hmmm....

steveklabnik:
So, this is strange. If I panic! on the final document, I get this https://gist.github.com/steveklabnik/afe548237ee00c8875b933fe2d296a57

But the JSON outputted is this? 

```json
{
  "data": {
    "type": "crate",
    "id": "example",
    "attributes": {
      "docs": " An example crate.\n\n This crate is used to test out the new rustdoc.\n\n It should have an example of almost everything, for testing\n purposes.\n"
    },
    "relationships": {
      "modules": {
        "data": [
          {
            "type": "module",
          }
        ]
      }
    }
  },
```

note that you can see another bug: the crate root has every child instead of just its own children. I expect this to be true for everything...

steveklabnik:
Eh I shouldn't have pushed those last three commits, resetting

steveklabnik:
I think I'm gonna leave this one till Monday, if anyone wants to poke at it, please do!

steveklabnik:
@mgattozzi what do you think about taking some time at rustconf to pair on this? 😄

mgattozzi:
Sounds good! I'll be landing in Portland around 10am or so and I'll be at the Nines and I only have the speaker dinner that night. I was also planning to meet up with @hjr3 who wanted to pair for rustdoc so why don't we plan to do so then?

mgattozzi:
Maybe after you're done with the beginner session.

steveklabnik:
This is a tough rebase and not much code; I will open up a new PR when I get to tackling it.

PR #110, opened 2017-08-03
Add phantomjs to package.json

### LINKED ISSUE #109: Fix the mac build
Looks like @mgattozzi 's https://github.com/steveklabnik/rustdoc/pull/107 broke the build. I assumed that it wouldn't and merged too quickly 😓 

https://travis-ci.org/steveklabnik/rustdoc/jobs/260764460

something about phantomjs

mgattozzi:
phantomjs is already supposed to be built into travis but it doesn't
seem like it is for OSX. This adds it as a dev dependency to unbork the
build.

Fixes #109


steveklabnik:
Thanks!

PR #112, opened 2017-08-04
Whitelist files in build.rs

### LINKED ISSUE #58: Improve build.rs to not add extra files
In https://github.com/steveklabnik/rustdoc/commit/718768e2c8ddb51331e244f0a7655b952bb3217f, we added a `build.rs` to properly package up the assets.

The problem is, I just spent about two hours debugging a problem that boiled down to "hey steve had a `data.json` file lying inside of `frontend/dist` for some reason. So my code was writing out a `data.json`, only to immediately be overwritten by the old file. Ugh!

We can't do a literal whitelist because of the hashes, but maybe blacklisting other files isn't good... I dunno.

hjr3:
build.rs would add every file in frontend/dist to asset.in. This would
cause hard to debug errors when files were unintentionally placed in
frontend/dist.

Fixes #58

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -5,7 +5,7 @@ extern crate error_chain;
 #[macro_use]
 extern crate quote;
 extern crate syn;
-extern crate walkdir;
+extern crate glob;
 
 use std::env;
 use std::fmt;
@@ -17,12 +17,13 @@ use std::process::exit;
 use std::thread;
 
 use quote::Ident;
-use walkdir::WalkDir;
+use glob::glob;
 
 error_chain! {
     foreign_links {
         Io(::std::io::Error);
-        WalkDir(::walkdir::Error);
+        PatternError(::glob::PatternError);
+        GlobError(::glob::GlobError);
     }
 }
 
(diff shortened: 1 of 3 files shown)


build.rs
```diff
@@ -65,14 +66,23 @@ fn main() {
...
-
-    for entry in WalkDir::new(DIST).into_iter() {
-        let entry = entry?;
-        if entry.metadata()?.is_file() {
-            output.push(Asset {
-                // If the directory isn't valid this wouldn't have worked.
-                path: String::from(entry.path().to_str().unwrap()).replace("\\", "/"),
-            });
+    let whitelist = vec![
+        "assets/*",
```
euclio: will this include subfolders?

hjr3: This particular pattern will not. If we write it as `assets/**` it will include subfolders. The glob pattern syntax is [here](https://doc.rust-lang.org/glob/glob/struct.Pattern.html).

mgattozzi:
@hjr3 looks like the tests are failing due to not being able to find files. I think you need to prepend "frontend/" to each of those paths in the whitelist.

build.rs
euclio: Ok, I think that we should include subfolders as well.

hjr3: fixed

hjr3:
I believe this is ready to go. The build is still failing due to upstream issues. Let me know if there is something else I should do.

euclio:
The build should pass with a rebase.

hjr3:
@euclio indeed it does now!

PR #153, opened 2017-08-11
Add an 'open' command

### LINKED ISSUE #90: Provide a subcommand to open the docs in the default browser
Bikeshedding the name:`rustdoc open`, `rustdoc run`, `rustdoc build --open`

~This might be blocked on #40...?~

mbrubeck:
Fixes #90.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -71,7 +71,7 @@ impl Config {
 
     /// Open the generated docs in a web browser.
     pub fn open_docs(&self) -> Result<()> {
-        open::that(self.output_path().join("index.html")).unwrap(); // TODO: Error handling
+        open::that(self.output_path().join("index.html"))?;
         Ok(())
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,6 +10,7 @@ extern crate serde_derive;
 extern crate serde_json;
 
 extern crate indicatif;
+extern crate open;
 extern crate rayon;
 extern crate rls_analysis as analysis;
 extern crate serde;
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -62,6 +63,17 @@ impl Config {
...
     }
+
+    /// Returns the directory where output files should be placed
+    pub fn output_path(&self) -> PathBuf {
+        self.manifest_path.join("target").join("doc")
+    }
+
+    /// Open the generated docs in a web browser.
+    pub fn open_docs(&self) -> Result<()> {
+        open::that(self.output_path().join("index.html")).unwrap(); // TODO: Error handling
```
steveklabnik: given that this returns Result already, `?` doesn't already work here?

mbrubeck: It doesn't return an io::Result, and I wasn't sure how to map its error type to an io::Error.

steveklabnik: I *think* because we use error-chain, you just add in the wrapper for its type and it should Just Work? I'm still learning error-chain

mbrubeck: Oops, I was just reading the code wrong. Fixed.

mgattozzi: @mbrubeck doesn't look like you actually put the change in for this at all since error.rs never got updated with the error type and the unwrap is still here.

mbrubeck: This comment is on an obsolete diff.  Also, I was doubly wrong; It *does* return an io::Error, so it nothing needed to be added to `error.rs`.

euclio:
I feel like this would be better suited as `rustdoc build --open`. Otherwise you might open stale docs.

mbrubeck:
This PR currently adds both `build --open` which builds and opens, and `open` which just opens.  I think these are both useful.  If the docs are already build and you just want to re-open them, it's handy to do so without triggering a rebuild.

As a special case if the docs have never been build, `open` will build them.  (The alternative would be to print an error saying "please run `build` first", which seemed strictly less useful.)

euclio:
I see, guess I didn't look at the code closely enough! Thanks for the explanation.

mbrubeck:
Fixed rustfmt errors.

Issue #155, opened 2017-08-11
Reconcile old rustdoc strategy with new


euclio:
I think there's a bit of an mismatch between the fact that old rustdoc allows documenting individual files while the new rustdoc expects a crate with Cargo.toml. How do we want to handle that?

Another way of putting it is that the old rustdoc expects to be wrapped by cargo, while the new rustdoc expects to wrap cargo.

cc #125

steveklabnik:
We talked about this at the tools team meeting today and @nrc is a fan of removing this functionality entirely. That is, we want to support writing markdown files as a way of writing long-form docs, (https://github.com/steveklabnik/rustdoc/issues/24) but this is kind of a low-level way of doing that, and should be treated as a more first-class thing.

euclio:
Are we talking about the same thing? I originally opened the issue intending to talk about individual *Rust* files, not markdown. You can invoke the old rustdoc like `rustdoc some_file.rs` and it will create documentation for that file, without needing a `Cargo.toml`. Not saying we need to support that use case, but I'd like us to make an official decision.

steveklabnik:
Oh!

Since that's just a crate root, yeah, okay. Let me change the title again; I misunderstood.

PR #156, opened 2017-08-13
Change --manifest-path to point to Cargo.toml.

### LINKED ISSUE #113: --manifest-path argument should point to Cargo.toml, not its directory
This would make it consistent with the cargo argument.

@steveklabnik, have you considered making an easy issue tag for new contributors? This would be a good candidate.

mjkillough:
Make the `--manifest-path` option consistent with Cargo's by having it
point to the `Cargo.toml`, rather than the directory containing the
`Cargo.toml`.

If the `--manifest-path` option is not provided, the default is still to
look for a `Cargo.toml` in the current working directory. This is
different behavior to Cargo, which attempts to look for a `Cargo.toml`
by traversing up through the current working directory and its parents.
A future enhancement could fix this, but it would make sense to wait for
the outcome of #155.

Fixes #113.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -17,7 +17,7 @@
             "args": [
                 "run",
                 "--",
-                "--manifest-path=example",
+                "--manifest-path=example/Cargo.toml",
                 "build"
             ]
         },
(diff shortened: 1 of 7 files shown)


steveklabnik:
This looks great, thank you so much!

PR #169, opened 2017-09-08
Upgrade error-chain to 0.11

### LINKED ISSUE #100: Upgrade to error-chain 0.11 when released
In https://github.com/steveklabnik/rustdoc/pull/99 we upgraded to a prerelease version of error-chain in order to suppress warnings. We should use the real 0.11 when it's out.

I'm tagging this as help-wanted because it's an easy bug, but it's not possible until they actually have a new release; I don't know how long that will be.

mgattozzi:
We had used a prerelease of this to get rid of some annoying warnings.
This moves the code over to the final 0.11 release.

Closes #100

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ doctest = false
 
 [dependencies]
 clap = "2.24.2"
-error-chain = "=0.11.0-rc.2"
+error-chain = "0.11"
 indicatif = "0.6.0"
 rayon = "0.8.2"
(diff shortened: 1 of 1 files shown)


steveklabnik:
Travis' mac situation is..... yeah. So I'm going to merge this anyway, thanks a ton @mgattozzi !

mgattozzi:
It's always like that unfortunately.

PR #171, opened 2017-09-19
Properly nest definitions

steveklabnik:
This is a resurrection of https://github.com/steveklabnik/rustdoc/pull/108, time to get it fixed 😄 

@mgattozzi had wanted to help out with this, so here's the failing tests! 

Basically, https://github.com/steveklabnik/rustdoc/blob/master/src/json/mod.rs#L16 needs to be modified to generate the proper structure. More specifically, https://github.com/steveklabnik/rustdoc/blob/master/src/json/mod.rs#L27 doesn't totally do its job; it assumes all things are flat.

### DIFF
--- a/src/json/api.rs
+++ b/src/json/api.rs
@@ -194,6 +194,32 @@ impl Document {
             }
         }
     }
+
+    /// Like relationships, but if the key already exists, appends data rather than
+    /// overwriting.
+    pub fn add_relationship(&mut self, ty: String, data: Data) {
+        match self.relationships {
+            None => {
+                let mut data_map = HashMap::with_capacity(DATA_SIZE);
+                data_map.insert(String::from("data"), vec![data]);
+
+                let mut relationships = HashMap::with_capacity(METADATA_SIZE);
+                relationships.insert(ty, data_map);
+                self.relationships = Some(relationships);
+            }
+            Some(ref mut relationships) => {
+                let entry = relationships.entry(ty).or_insert_with(|| {
+                    let mut map = HashMap::with_capacity(DATA_SIZE);
+                    map.insert(String::from("data"), Vec::new());
+                    map
+                });
+
+                // We know that we have a "data" entry because we created it above
+                let data_vec = entry.get_mut("data").unwrap();
+                data_vec.push(data);
+            }
+        }
+    }
 }
 
 impl Data {
--- a/src/json/mod.rs
+++ b/src/json/mod.rs
@@ -4,16 +4,23 @@ mod api;
 
 pub use self::api::*;
 
-use std::collections::HashMap;
-
-use analysis::{self, AnalysisHost, DefKind};
-use rayon::prelude::*;
+use analysis::{AnalysisHost, DefKind};
 use serde_json;
 
 use error::*;
 
+use std::collections::VecDeque;
+
 /// This creates the JSON documentation from the given `AnalysisHost`.
 pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
+    // This function does a lot, so here's the plan:
+    //
+    // First, we need to process the root def and get its list of children.
+    // Then, we process all of the children. Children may produce more children
+    // to be processed too. Once we've processed them all, we're done.
+
+    // Step one: we need to get all of the "def roots", and then find the
+    // one that's our crate.
     let roots = host.def_roots()?;
 
     let id = roots.iter().find(|&&(_, ref name)| name == crate_name);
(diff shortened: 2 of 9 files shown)


mgattozzi:
As we discussed at RustConf it's basically getting these into a Tree and when inserted takes it's parent node as part of it's module path. Then just iterate over the collection like normal to put it in the JSON as a flat structure. Once we get this done we can start cruising through things

steveklabnik:
I've pushed a commit that's *almost* there but not quite; I'm more about keeping things compiling than making it correct just yet. I believe in this version, the relationships is too global, and so it's going to be wrong, but that's fine for now.

I have since realized two things:

1. we don't need to do this recursively
2. the tree idea @mgattozzi references above doesn't matter

#### We don't need to do this recursively

channels already give us the needed semantics. Consider this: https://play.rust-lang.org/?gist=736b1257468e97f2c9fdcba91cb24f5b&version=stable

It's like a work queue, we pop stuff off, work on it, and then maybe push stuff on, maybe not. Once the queue is empty, we're done! This saves a *massive* amount of lifetime headaches.

It's not parallel, hence the removal of rayon, but it shouldn't be too painful to bring back in the future, as this is basically what rayon does internally.


#### no tree needed

In the end, we start off with a list of things, and we need the included array at the end. Both are basically flat lists. Yes, it's true there's a tree inside, but we were only going to go with the tree because, in the recursive solution, we were walking the tree.  Now that we're not really doing a classic recursive tree-walk, there's no need to introduce a tree in the middle: we simply process each thing individually until we're done.

---------------------------------

I'm going to leave this here and enjoy my Friday evening, I'm not likely to pick it up again until Monday. If anyone wants to try to fix the remaining bugs here with relationships, that'd be awesome! Otherwise, I'll try to polish it off then.

mgattozzi:
This gives context to your tweet. Niiiice. I might take a look at this but in all reality I might just leave if for you on Monday :D

steveklabnik:
The commits I've just pushed pass locally 🎊 

please check out the diff; I feel like we could make this code better, but I'm pushing this up now while it actually works 😉

steveklabnik:
```text
Rustfmt failed at /home/travis/build/steveklabnik/rustdoc/tests/source/nested_modules.rs:8: line exceeded maximum length (maximum: 100, found: 111) (sorry)
```

Maybe we should stop caring about rustfmt in tests, as we can't break up these lines and they're gonna bump up against this limit...

steveklabnik:
I pushed a commit to do that, but would be interested in what @mgattozzi and @euclio think

tests/source.rs
```diff
@@ -29,7 +29,6 @@ extern crate regex;
 extern crate rls_analysis as analysis;
 extern crate shlex;
 extern crate tempdir;
-extern crate jsonapi;
```
steveklabnik: this change is technically unrelated to this PR, but was causing a warning so I tossed it in here.

euclio:
I'm 100% on board unblocking CI on rustfmt, but I think it's nice to print out a diff.

mgattozzi:
I hate rustfmt so I'm okay with it unblocking. A diff would be good.

tests/source.rs
mgattozzi: Might as well!

src/json/api.rs
```diff
@@ -194,6 +194,35 @@ impl Document {
             }
         }
     }
+
+    /// like relationships, but if the key already exists, appends data rather than
```
mgattozzi: s/like/Like/g

src/json/api.rs
```diff
@@ -194,6 +194,35 @@ impl Document {
...
+                let mut data_map = HashMap::with_capacity(DATA_SIZE);
+                data_map.insert(String::from("data"), vec![data]);
+
+                let mut relationships = HashMap::with_capacity(METADATA_SIZE);
+                relationships.insert(ty, data_map);
+                self.relationships = Some(relationships);
+            }
+            Some(ref mut relationships) => {
+                //let mut data_map = HashMap::with_capacity(DATA_SIZE);
+                //data_map.insert(String::from("data"), data);
```
mgattozzi: Stray comments

src/json/api.rs
```diff
@@ -194,6 +194,35 @@ impl Document {
...
+
+                let entry = relationships.entry(ty).or_insert_with(|| {
+                    let mut map = HashMap::with_capacity(DATA_SIZE);
+                    map.insert(String::from("data"), Vec::new());
+                    map
+                });
+
+                // we know that we have a "data" entry because we created it above
```
mgattozzi: s/we/We

steveklabnik:
Cool, I *think* all I have to do is remove `--write-format diff`. Let's see!

src/json/mod.rs
```diff
@@ -24,63 +31,88 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
...
-            .map(|id: analysis::Id| recur(&id, host))
-            .reduce(Vec::default, |mut a: Vec<analysis::Def>,
-             b: Vec<analysis::Def>| {
-                a.extend(b);
-                a
-            });
-        defs.extend(child_defs);
-        defs
-    }
+    // create the main `Document`
```
mgattozzi: s/c/C

src/json/mod.rs
```diff
@@ -24,63 +31,88 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
...
+        .id(crate_name.to_string())
+        .attributes(String::from("docs"), root_def.docs);
 
-    let mut included: Vec<Document> = Vec::new();
-    let mut relationships: HashMap<String, Vec<Data>> = HashMap::with_capacity(METADATA_SIZE);
+    // Now that we have that, it's time to get the children; these are
+    // the top-level items for the crate.
+    let ids = host.for_each_child_def(root_id, |id, _def| id).unwrap();
+
+
```
mgattozzi: Extra line

src/json/mod.rs
```diff
@@ -24,63 +31,88 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
...
+    // as a work queue; we take an item off, process it, and then if it has
+    // children, push them onto the queue. When the queue is empty, we've processed
+    // everything.
+    //
+    // Additionally, we generate relationships between the crate itself and
+    // these ids, as they're at the top level and hence linked with the crate.
 
-    for def in recur(&root_id, host) {
-        let (ty, relations_key) = match def.kind {
+    let (sender, receiver) = mpsc::channel();
```
mgattozzi: Clever! mpsc was probably the better choice but we were both stuck on recursion to not notice it 😅

src/json/mod.rs
```diff
@@ -24,63 +31,88 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
...
             DefKind::Struct => (String::from("struct"), String::from("structs")),
             _ => continue,
         };
 
+        let data = Data::new().ty(ty.clone()).id(def.qualname.clone());
+
+        document.add_relationship(child_ty, data);
+    }
+
+    // the loop below is basically creating this vector
```
mgattozzi: s/t/T

src/json/mod.rs
```diff
@@ -24,63 +31,88 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
...
+    let mut included: Vec<Document> = Vec::new();
+
+    while let Ok(id) = receiver.try_recv() {
+        // push each child to be processed itself, and also record
```
mgattozzi: s/push/Push

src/json/mod.rs
```diff
@@ -24,63 +31,88 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
...
+        // their ids so we can create the relationships for later
+        let child_ids = host.for_each_child_def(id, |id, _def| {
+            sender.send(id).unwrap();
+            id
+        })?;
+
+        // process this one
```
mgattozzi: Is this a stray comment?

src/json/mod.rs
```diff
@@ -24,63 +31,88 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
...
+
+        // question: we could do this by cloning it in the call to for_each_child_def
```
mgattozzi: s/question/Question

As for the answer we'd need some kind of benchmark. I think we don't need to worry about this too much for now and should focus more on getting things working then optimization.

src/json/mod.rs
```diff
@@ -24,63 +31,88 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
...
+        // above/below; is that cheaper, or is this cheaper?
+        let def = host.get_def(id).unwrap();
+
+
```
mgattozzi: Stray line

src/json/mod.rs
```diff
@@ -24,63 +31,88 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
...
-                .attributes(String::from("name"), def.name)
-                .attributes(String::from("docs"), def.docs),
-        );
-
-        let item_relationships = relationships.entry(relations_key).or_insert_with(
-            Default::default,
-        );
-        item_relationships.push(Data::new().ty(ty).id(def.qualname));
-    }
+        let (ty, child_ty) = match def.kind {
```
mgattozzi: Would it be possible to have `String` consts at all? We have a lot of these lying around.

src/json/mod.rs
```diff
@@ -24,63 +31,88 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
...
+
+            let data = Data::new().ty(ty.clone()).id(def.qualname.clone());
+
+            document.add_relationship(child_ty.clone(), data);
+        }
 
-    // Insert all of the different types of relationships into this `Document` type only
-    for (ty, data) in relationships {
-        data_document.relationships(ty, data);
+        // put the document into our included vector
```
mgattozzi: s/put/Put

src/json/mod.rs
euclio: Could we use `VecDeque::push_back` and `VecDeque::pop_front` instead?

steveklabnik: @euclio what's the benefit there?

I'm also interested in re-introducing Rayon... at some point

src/json/mod.rs
steveklabnik: Yeah so this is a good question; the second set also differs slightly.

Also, these may want to be literals and call the .to_string below instead, idk

src/json/mod.rs
steveklabnik: updated to use VecDeque

steveklabnik:
I'm going to clean up these bits tomorrow and then :shipit: , let me know if, other than the stray comments, there are other blockers!

src/json/mod.rs
mgattozzi: We have a lot of `String`s in use in other places who's only thing is to be added as a key. It'd probably be faster to have consts for use throughout. Not a huge thing and can be addressed later. This isn't a blocker.

mgattozzi:
@steveklabnik I don't see any reason to block. Let's let Travis run it's course and go from there unless @euclio spots anything I might have missed.

mgattozzi:
@steveklabnik yay LGTM now!

src/json/mod.rs
euclio: Thanks. I was thinking that if we are only using a channel for its queue-like properties, we might as well just use a queue for a performance boost. Plus, it might be confusing to see it being used in a non-concurrent context.

Issue #158, opened 2017-08-14
Run Appveyor/Travis on a daily basis


mgattozzi:
Travis and Appveyor both have the ability to schedule builds. Since we use nightly it would be great to set them up to run nightly as well to check for possible breakage from newer versions of Rust.

steveklabnik:
👍

mgattozzi:
@steveklabnik since you're the only one with access to those you'll have to be the one to set it up

steveklabnik:
I've updated travis to do it; i had to email appveyor support to ask about recurring builds, we'll see what they say!

steveklabnik:
appveyor is now turned on too!

mgattozzi:
Yay!

PR #174, opened 2017-09-20
Update build scripts to update deps

mgattozzi:
Cargo.lock is part of the build which is fine for reproducible builds.
However, with the use of nightly and unstable features in things like
rls-analysis it means things can break and already have quite a bit.
This commit relaxes the constraints on rls-analysis to use the newest
version (which is good so we can identify problems later) and adds
a `cargo update` to the CI builds to avoid problems with using older
cached versions of rls-analysis.

cc/ @steveklabnik


steveklabnik:
@nrc what do you think of this change? I am of two minds.

mgattozzi:
I'd also be fine with making the update to the dep version just so we can fix the build. Either works.

nrc:
I feel like I'm missing something. Won't this make you break more often? Because you'll randomly pull in every upstream change, some of which might be breaking? You still get breakage due to nightly changes until that is fixed upstream, so it seems the only advantage is that once something is fixed upstream, you get the fix quicker?

mgattozzi:
Yeah my wording was incorrect I think. It's mostly we get upstream fixes quicker. We could just update the dep and have Travis run a nightly Cron job that will ping at least @steveklabnik if there's a breaking change instead.

steveklabnik:
That is https://github.com/steveklabnik/rustdoc/issues/158, which I thought I fixed, but did not.

@mgattozzi yeah, let's go with that. Let's change up this PR to only update the dep, and I will bump #158 up in priority for me 😄

mgattozzi:
@steveklabnik I've updated the PR and changed the commit to better reflect the change.

steveklabnik:
Thanks!

Issue #121, opened 2017-08-08
Support non-module items: backend


steveklabnik:
Once I work the bugs out of https://github.com/steveklabnik/rustdoc/pull/108, it'll be time to support the other kinds of items.

Fundamentally, what needs to happen is with [this line](https://github.com/steveklabnik/rustdoc/blob/782fd7f8c5e6cef878183c52121ef5414872e0c5/src/json/mod.rs#L52); instead of `_ => continue`, it should have every [`DefKind`](https://docs.rs/rls-analysis/0.6.5/rls_analysis/enum.DefKind.html).

- [ ] Enum
- [ ] Tuple
- [ ] Struct
- [ ] Union
- [ ] Trait
- [ ] Function
- [ ] Method
- [ ] Macro
- [ ] Mod
- [ ] Type
- [ ] Local
- [ ] Static
- [ ] Const
- [ ] Field

I am not actually sure about `Local`, as I'd imagine local items don't need to be documented.

mgattozzi:
We just finished what #108 was trying to do with #171. Time to start working on this issue now. Definitely a good place to start if you're new to the project/have some Rust experience. Willing to help mentor as well for these.

projektir:
I'm assuming `ty` stands for "type"?

How does one determine what should be returned for every DefKind here? There's more to it than strings like "module", "modules", right?

steveklabnik:
Yeah, `ty` is type.

> How does one determine what should be returned for every DefKind here? There's more to it than strings like "module", "modules", right?

For the backend, it needs to be in two places; https://github.com/steveklabnik/rustdoc/blob/48794bde699955b38c4ea849d915d1e4cfd510c2/src/json/mod.rs#L60 and https://github.com/steveklabnik/rustdoc/blob/48794bde699955b38c4ea849d915d1e4cfd510c2/src/json/mod.rs#L88 note the small differences. That's because....

for the frontend, https://github.com/steveklabnik/rustdoc/tree/48794bde699955b38c4ea849d915d1e4cfd510c2/frontend/app/models 

[...]

Issue #7, opened 2017-07-11
Does `rls-analysis` support everything we need?


steveklabnik:
Upstream wants to work with us, but we haven't actually determined if we need anything yet.

This quest issue is both broad and specific. Someone wanting to close it out should:

1. Work through all of the different kinds of information that we need out of `rls-analysis`.
2. Inspect its API to figure out if it all exists or not.

If we don't have all of the stuff we need, we'll have to open issues with upstream and work with them to get the information landed.

dzamlo:
After a quick look at generated JSON, it looks like the exported macros (and their doc comments) are not included.

But I'm not familiar enough with the `save-analysis` part of rust to be sure if this is a mistake on my part or if this is really not included.

euclio:
@steveklabnik It seems that rls-analysis does not export the visibility of items? We shouldn't be documenting private items :)

steveklabnik:
yeah, we should at least know if something is public or private; sounds like an important thing to sort out

theindigamer:
@euclio , it would be nice to be able to document private definitions too. Earlier discussion https://github.com/rust-lang/rust/issues/15347

mgattozzi:
We ran into some issues implementing the DefKinds. The discussion was [here](https://github.com/steveklabnik/rustdoc/pull/177#discussion_r141922958). Essentially, rls-analysis still needs to be expanded for other items so we can implement everything.

euclio:
We will not be able to support `pub use` and reexports until https://github.com/nrc/rls-analysis/issues/123 is fixed.

PR #177, opened 2017-09-27
Adding all the DefKinds #121

projektir:
Addresses #121.

This excludes `Local`, which means it needs to be deliberately skipped. Maybe it should be more specific like `DefKind::Local => continue` in case something else is added in the future?

I'm not sure if it makes sense for `Static` or `Const` and such to have children?

I believe the idea is to set this boilerplate up and then we can move on to actually implementing things in the frontend?

### DIFF
--- a/example/src/lib.rs
+++ b/example/src/lib.rs
@@ -19,9 +19,48 @@ impl ExampleStruct{
     }
 }
 
+/// An enum.
+pub enum SampleEnum {
+    /// An enum variant.
+    EnumVariant,
+}
+
+/// An example function.
+///
+/// This function does stuff!
 pub fn sample_function(x: i32) -> ExampleStruct {
     let _ = x;
     ExampleStruct
+}
+
+/// A type ExampleType.
+type ExampleType = String;
+
+/// An example macro.
+macro_rules! example_macro {
+    () => ( println!("Hello!"); )
+}
+
+/// An example union.
+union example_union {
+    f1: u32,
+    f2: f32,
+}
+
+/// A const.
+const EXAMPLE_CONST: i32 = 5;
+
+/// A static.
+static EXAMPLE_STATIC: i32 = 5;
+
+/// A struct that contains another struct and other fields.
+///
+/// Docs for the ContainerStruct.
+pub struct ContainerStruct {
+    /// These are integer field docs.
+    integer: i32,
+    /// These are docs for the inner struct.
+    inner_struct: ExampleStruct,
 }
 
 /// An example module
--- a/src/json/mod.rs
+++ b/src/json/mod.rs
@@ -61,6 +61,19 @@ pub fn create_documentation(host: &AnalysisHost, crate_name: &str) -> Result<Doc
         let (ty, child_ty) = match def.kind {
             DefKind::Mod => (String::from("module"), String::from("modules")),
             DefKind::Struct => (String::from("struct"), String::from("structs")),
+            DefKind::Enum => (String::from("enum"), String::from("enums")),
+            DefKind::Trait => (String::from("trait"), String::from("traits")),
+            DefKind::Function => (String::from("function"), String::from("functions")),
+            DefKind::Type => (String::from("type"), String::from("types")),
+            DefKind::Static => (String::from("static"), String::from("statics")),
+            DefKind::Const => (String::from("const"), String::from("consts")),
+            DefKind::Field => (String::from("field"), String::from("fields")),
+            DefKind::Tuple => continue,
+            DefKind::Local => continue,
+            // The below DefKinds are not supported in rls-analysis
+            // DefKind::Union => (String::from("union"), String::from("unions")),
+            // DefKind::Macro => (String::from("macro"), String::from("macros")),
+            // DefKind::Method => (String::from("method"), String::from("methods")),
             _ => continue,
         };
 
@@ -86,9 +99,22 @@ pub fn create_documentation(host: &AnalysisHost, crate_name: &str) -> Result<Doc
 
         // Using the item's metadata we create a new `Document` type to be put in the eventual
         // serialized JSON.
-        let (ty, child_ty) = match def.kind {
-            DefKind::Mod => (String::from("module"), String::from("child_modules")),
-            DefKind::Struct => (String::from("struct"), String::from("child_structs")),
+        let ty = match def.kind {
+            DefKind::Mod => String::from("module"),
+            DefKind::Struct => String::from("struct"),
+            DefKind::Enum => String::from("enum"),
+            DefKind::Trait => String::from("trait"),
+            DefKind::Function => String::from("function"),
+            DefKind::Type => String::from("type"),
+            DefKind::Static => String::from("static"),
+            DefKind::Const => String::from("const"),
+            DefKind::Field => String::from("field"),
+            DefKind::Tuple => continue,
+            DefKind::Local => continue,
+            // The below DefKinds are not supported in rls-analysis
+            // DefKind::Union => (String::from("union"), String::from("unions")),
+            // DefKind::Macro => (String::from("macro"), String::from("macros")),
+            // DefKind::Method => (String::from("method"), String::from("methods")),
             _ => continue,
         };
 
@@ -100,6 +126,24 @@ pub fn create_documentation(host: &AnalysisHost, crate_name: &str) -> Result<Doc
 
         for id in child_ids {
             let def = host.get_def(id).unwrap();
+            let (ty, child_ty) = match def.kind {
+                DefKind::Mod => (String::from("module"), String::from("child_modules")),
+                DefKind::Struct => (String::from("struct"), String::from("child_structs")),
+                DefKind::Enum => (String::from("enum"), String::from("child_enums")),
+                DefKind::Trait => (String::from("trait"), String::from("child_traits")),
+                DefKind::Function => (String::from("function"), String::from("child_functions")),
+                DefKind::Type => (String::from("type"), String::from("child_types")),
+                DefKind::Static => (String::from("static"), String::from("child_statics")),
+                DefKind::Const => (String::from("const"), String::from("child_consts")),
+                DefKind::Field => (String::from("field"), String::from("child_fields")),
+                DefKind::Tuple => continue,
+                DefKind::Local => continue,
+                // The below DefKinds are not supported in rls-analysis
+                // DefKind::Union => (String::from("union"), String::from("unions")),
+                // DefKind::Macro => (String::from("macro"), String::from("macros")),
+                // DefKind::Method => (String::from("method"), String::from("methods")),
+                _ => continue,
+            };
 
             let data = Data::new().ty(ty.clone()).id(def.qualname.clone());
 
--- a/tests/source/enums.rs
+++ b/tests/source/enums.rs
@@ -1,0 +1,19 @@
+#![crate_type = "lib"]
+
+// @has /data/relationships/enums/data/0/type 'enum'
+// @has /data/relationships/enums/data/0/id 'enums::SampleEnum'
+
+// @has /included/0/type 'enum'
+// @has /included/0/id 'enums::SampleEnum'
+// @has /included/0/attributes/name 'SampleEnum'
+// @has /included/0/attributes/docs 'An enum.'
+
+// variants do not work yet, see https://github.com/nrc/rls-analysis/issues/98
+// has /included/0/relationships/child_enums/data/0/type 'enum'
+// has /included/0/relationships/child_enums/data/0/id 'enums::SampleEnum::EnumVariant'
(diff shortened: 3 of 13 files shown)


steveklabnik:
> Maybe it should be more specific like DefKind::Local => continue in case something else is added in the future?

Yeah, seems good.

> I'm not sure if it makes sense for Static or Const and such to have children?

Yeah, this is true. They shouldn't have children, they probably *will be* children, though.

Could you add tests please?

mgattozzi:
For the tests you would want to put something under this directory [here](https://github.com/steveklabnik/rustdoc/tree/master/tests/source). I think something like def_kind.rs would be a good thing and then just create every type of `DefKind` in the source code there and set up some assertions. Or seperately (and probably a better idea) is to set up one type of file for each DefKind we don't have yet and make the assertions in there. You can look at the other files for examples of how to assert something is of the output JSON that rustdoc emits.

@projektir does that make sense? Anything I can clear up with my explanation?

projektir:
@mgattozzi 
OK, I've tried some of this but I'm hitting a lot of issues (see comments)... this commit will break since otherwise I have to delete the whole file for macros.

tests/source/macros.rs
```diff
@@ -0,0 +1,9 @@
+#![crate_type = "lib"]
+
+// @has /data/relationships/macros/data/0 'macro'
```
projektir: I can't get this to pick up at all. It doesn't seem to see macros. I have the same issue with unions and I imagine would have the same issue with statics and consts.

tests/source/methods.rs
```diff
@@ -0,0 +1,13 @@
...
+// @has /data/relationships/structs/data/0/id 'methods::UnitStruct'
+
+// @has /data/relationships/methods/data/0/id 'method'
+
+// has /included/0/child_methods 'method'
+
+pub struct UnitStruct;
+
+impl UnitStruct {
+    pub fn method_function() {}
```
projektir: Since to create a method it needs to be attached to some object, I need to identify it somehow, but nothing seems to work. Am I misunderstanding that this is not actually a child of a struct or such? Is there some intermediate type?

projektir:
Looking at it a bit more, it seems the `data.json` (which I am assuming is backend relevant) is not getting generated right. It only has modules, structs, and functions, and doesn't get methods nor other things if I modify the `lib.rs`. So it looks like my boilerplate code is just not getting the job done in the first place.

tests/source/methods.rs
steveklabnik: @nrc, I don't know anything about this, maybe you do?

mgattozzi:
Yeah it might just something with rls-analysis or something in your code (which I doubt).

tests/source/methods.rs
nrc: Yeah, basically nothing with impls (and thus methods) is going to work very well yet. We need some work in rustc to add support and that is a bit non-trivial in the general case. We could add support for inherent impls pretty easily, if someone wanted to hack on that I could give instructions.

nrc: And actually, I thought inherent impls worked, but maybe I was just wrong about that.

mgattozzi: In which case it might just be best for now if we put tests for what works, mark what doesn't (but still keep the code there), and have the tests written for whatever works. Thoughts @steveklabnik?

steveklabnik: I'd remove the code for the stuff that doesn't work, and go with a `_ =>`. But yeah, otherwise, let's do that, @projektir . It might also be worth noting this over at https://github.com/steveklabnik/rustdoc/issues/7

Thanks @nrc

tests/source/macros.rs
projektir: @steveklabnik what about this problem?

I'm seeing in #7 a comment"After a quick look at generated JSON, it looks like the exported macros (and their doc comments) are not included.". So that kind of seems like a problem. And I've also seen the same issue with unions.

I'm going to tweak the `lib.rs` file with all these things and see what it actually generates because it doesn't seem like a lot right now.

steveklabnik: sounds like maybe the same thing as impls, I'm not sure.

projektir:
All right, so I added everything that I could figure out. I'd like to say that the following DefKinds can be marked as done for  #121: `Trait`, `Function`, `Mod`, `Type`, `Static`, `Const`.
`Tuple` and `Local` should just be removed maybe since they won't be implemented? For the rest there seem to be some bugs.

I've also added various examples to the example `lib.rs`.

The new `lib.rs` contains the following: https://gist.github.com/projektir/5f51fc77fe0608e06f3d374d1b1bce03

Here's the resulting `data.json`: https://gist.github.com/projektir/f4070a06cb88b8e8093cdfd2f54dd24b

Seeing a few weird things:

1. Absence of `Macros` and `Unions` as we already know.
2. Can't see the impl and therefore `Methods` as we already know.
3. Information seems to get duplicated for `Structs`, struct `Fields`, and `Enums`. If you look you'll see EnumVariant is listed twice, etc. This doesn't happen for modules...
4. When `child_structs` and such is used, it seems that it'd be more accurate to say `struct_children`... or just plain `children`... but this is odd in general because while the `ContainerStruct` has as a field that's _not_ a struct, its type is still marked as struct (see comment in test case). That doesn't seem right. Similarly, enum variants have type `enum`, I think this is a related thing.
5. There's no way to see docs for EnumVariants. It seems to be because there's no `docs` attribute for children, only for top level includes, so I am not sure if EnumVariants need to be listed as includes or what.
6. Docs representation. Multiple lines of docs seem to go poof, only the first line is in the `.json`. Doc lines sometimes include spaces or `\n`, but this may be a problem with my IDE or something.

example/src/lib.rs
```diff
@@ -19,11 +19,50 @@ impl ExampleStruct{
     }
 }
 
+/// An enum.
+pub enum SampleEnum {
+    /// An enum variant.
```
projektir: No way to see these docs that I can find.

tests/source/fields.rs
```diff
@@ -0,0 +1,25 @@
...
+// @has /data/relationships/structs/data/0/type 'struct'
+// @has /data/relationships/structs/data/0/id 'fields::StructWithFields'
+
+// @has /included/0/type 'struct'
+// @has /included/0/id 'fields::StructWithFields'
+// @has /included/0/attributes/name 'StructWithFields'
+// @has /included/0/attributes/docs ''
+
+// This will fail, the type of 'fields::StructWithFields::integer' is 'struct'
+// has /included/0/relationships/child_structs/data/0/type 'field'
```
projektir: I expect this to be a `field`, but it's a `struct`. It seems "children" of something all get set to have the type as the parent, which of course is not true for complex types like Structs or Enums.

tests/source/structs.rs
```diff
@@ -1,10 +1,31 @@
...
+/// A unit struct.
+pub struct UnitStruct;
+
+// @has /data/relationships/structs/data/1/type 'struct'
+// @has /data/relationships/structs/data/1/id 'structs::ContainerStruct'
+
+// @has /included/1/type 'struct'
+// @has /included/1/id 'structs::ContainerStruct'
+// @has /included/1/attributes/name 'ContainerStruct'
+// @has /included/1/attributes/docs 'A struct that contains another struct.'
```
projektir: As you can see there's no "Docs for the ContainerStruct." here. That part of the documentation just seems to disappear.

tests/source/enums.rs
```diff
@@ -0,0 +1,18 @@
...
+// @has /included/0/id 'enums::SampleEnum'
+// @has /included/0/attributes/name 'SampleEnum'
+// @has /included/0/attributes/docs 'An enum.'
+
+// @has /included/0/relationships/child_enums/data/0/type 'enum'
+// @has /included/0/relationships/child_enums/data/0/id 'enums::SampleEnum::EnumVariant'
+
+/// An enum.
+pub enum SampleEnum {
+    /// An enum variant.
```
projektir: No way to see these docs that I can find.

steveklabnik:
> When child_structs and such is used, it seems that it'd be more accurate to say struct_children... or just plain children... but this is odd in general because while the ContainerStruct has as a field that's not a struct, its type is still marked as struct (see comment in test case). That doesn't seem right. Similarly, enum variants have type enum, I think this is a related thing.

Yeah, so I was doing the `child` thing because modules have children, but other types don't, so I'm wondering if this isn't, in the end, an anti-pattern. Hrm.

tests/source/fields.rs
steveklabnik: Hm, that seems... less than ideal.

nrc:
I think that I would avoid surfacing 'children' to the user at all - it's a bit vague and doesn't reflect any term in the code. I would use different terms for different structures - variants of an enum, fields of a struct, etc. Internally, it makes sense to use children, at least for save-analysis where we try to abstract things as much as possible.

It seems like there are a few rough edges with the save-analysis data. I think they should mostly be easy to fix (unions, docs for enum variants, etc.). If you could file issues on the rls-analysis repo it is easier for me than the Rust repo (just too many notifications). I'll try and fix them or organise an impl period/mentoring effort around them when I get back from parental leave.

projektir:
@nrc 
I've made some issues for the stuff that I found. Note that my understanding of `rls-analysis` is limited, so I may be looking at the wrong thing somewhere.

I'd be happy to try to work on those issues, as well.

nrc:
@projektir thanks for filing the issues! It would be great to have you work on some of them. I'm currently on parental leave, but I'll write some instructions when I get back in a couple of weeks.

mgattozzi:
@nrc if there's instructions left I can take a crack at them as well so we can get this PR moving.

steveklabnik:
Question: since this has been a while, do we want to merge, with the understanding that some stuff doesn't work yet, rather than block this on upstream work?

src/json/mod.rs
```diff
@@ -88,6 +101,19 @@ pub fn create_json(host: &AnalysisHost, crate_name: &str) -> Result<String> {
         let (ty, child_ty) = match def.kind {
             DefKind::Mod => (String::from("module"), String::from("child_modules")),
             DefKind::Struct => (String::from("struct"), String::from("child_structs")),
+            DefKind::Enum => (String::from("enum"), String::from("child_enums")),
```
steveklabnik: yeah, there's no real reason that we need the `child` stuff here; i was cargo-culting from modules

steveklabnik: this also means that this block and the one above are identical except for modules, which is annoying

steveklabnik:
Okay, so more thorough review today; we should cut the stuff that doesn't work yet, and leave in the stuff that does. wrt the child stuff, let's make it so that only `modules` does the child thing. And then we can ship this. Sound good?

steveklabnik:
Appveyor is failing on traits, but not for me or @projektir locally :/

steveklabnik:
Update: needed to rebase the branch; appveyor basically does this. Fixed that up. Now to figure out enums...

steveklabnik:
🎊

nrc:
@projektir Hey, are you still interested in working on some rls-analysis issues? I'm back from leave now, so you can ping me anytime in #rust-dev-tools or @rustc.

Some basic instructions for how to tackle these issues:
* I track the issues on the rls-analysis repo (you've already filed some issues there)
* the first job is figuring out where in the toolchain the problem is. There are only really three options:
  - rls-analysis - we get correct data from the compiler but then screw things up when processing the data. This is actually fairly rare. When it does happen, 9 times out of 10 you need to look in lowering.rs. If you log the info! statements, you should get a good idea of what is happening.
  - librustc_save_analysis - conceptually this is the last step of the compiler and it processes the compiler's data into the save-analysis JSON. Most bugs are here.
  - spans - sometimes save-analysis is working properly but the input is wrong. In this case the bug is usually in the parser and it is causing save-analysis to skip an expression or get the analysis wrong when it shouldn't.
* Write a small test case, then look at the JSON emitted by the compiler when run with `-Zsave-analysis`. You should be able to interpret the defs and refs to work out if the correct info is there. If it is, then the bug is in rls-analysis, if it isn't then the bug is in the compiler. Note that a ref without a def is ignored, so even if there is a ref, if it doesn't point at any def, it may as well be missing.
* If the bug is in the compiler, then look in dump_visitor. It is a visitor which walks the AST, so for your minimal test case you should be able to see where we ought to be dumping the node. Then use a debugger or write printlns to work out what is going on.
* Bugs with macros (or other generated code) tend to be 10x harder to solve, I'd ignore them to start with.

nrc:
previous comment also for @mgattozzi

mgattozzi:
@nrc I need to find some time hopefully soon I've been busier than expected but if I can I'd like to take a crack at it. I'll ping you whenever I have the time.

Issue #179, opened 2017-10-01
README is out of date


steveklabnik:
`DocData` no longer exists, so we should re-work the README to actually explain what's going on now. @projektir noticed that this got of date; this is my fault :(

mgattozzi:
There's a couple things it could use there beyond this as well. I might take a crack at this today.

PR #180, opened 2017-10-01
Update README.md

mgattozzi:
Our README was out of date. This updates it to have the current state
of the project reflected. It also cleans up the links so that it doesn't
pollute the reading of the text and cleans up a few other spellings,
switches '\' paths to '/' as both Windows and Unix systems can handle
'/' but both can't handle '\' which can be confusing to some users if
they just copy and paste the paths.

Closes: #179


mgattozzi:
@steveklabnik awesome glad you liked them!

PR #182, opened 2017-10-02
WIP: Refactor cargo commands

mackwic:
I push this PR now for easier review + early discussions.

Todo:
- [x] extract the command building into their own testable module 
- [x] create a `Command` bridge that we can inspect before calling the true `Command` builder
- [ ] there is probably an opportunity to have a `utils` module, used by the cargo module
- [ ] remove dead code
- [ ] ?

EDIT: Oh I forgot to add my new nick I got from the RustFest:

Sincerly,
The Test Dude

### DIFF
--- a/src/cargo/command_bridge.rs
+++ b/src/cargo/command_bridge.rs
@@ -21,7 +21,7 @@ impl CommandBridge {
     pub fn new(command_name: &str) -> Self {
         CommandBridge {
             command_name: String::from(command_name),
-            .. Default::default()
+            ..Default::default()
         }
     }
 
@@ -40,7 +40,7 @@ impl CommandBridge {
         value_str.push(value.as_ref());
 
         self.env.insert(key_str, value_str);
-        
+
         self
     }
 
@@ -83,8 +83,7 @@ mod tests {
             // arrange
             let arg = OsString::from("some arg");
             // act
-            let cmd = CommandBridge::new("some cmd")
-                .arg(&arg);
+            let cmd = CommandBridge::new("some cmd").arg(&arg);
             // assert
             assert!(cmd.args.contains(&arg))
         }
@@ -99,8 +98,7 @@ mod tests {
             let key = OsString::from("hello");
             let value = OsString::from("world");
             // act
-            let cmd = CommandBridge::new("plop")
-                .env(&key, &value);
+            let cmd = CommandBridge::new("plop").env(&key, &value);
             // assert
             assert_eq!(&value, cmd.env.get(&key).unwrap())
         }
@@ -114,8 +112,7 @@ mod tests {
             // arrange
             let stdout = Stdio::null();
             // act
-            let _cmd = CommandBridge::new("plop")
-                .stdout(stdout);
+            let _cmd = CommandBridge::new("plop").stdout(stdout);
             // assert
             // FIXME: is there is really no way to compare an Stdio instance ??
             //assert_eq!(stdout as *const (), cmd.stdout.unwrap() as *const ())
@@ -130,12 +127,10 @@ mod tests {
             // arrange
             let stderr = Stdio::null();
             // act
-            let _cmd = CommandBridge::new("plop")
-                .stderr(stderr);
+            let _cmd = CommandBridge::new("plop").stderr(stderr);
             // assert
             // FIXME: is there is really no way to compare an Stdio instance ??
             //assert_eq!(stderr as *const (), cmd.stderr.unwrap() as *const ())
         }
     }
 }
-
--- a/src/cargo/command_bridge.rs
+++ b/src/cargo/command_bridge.rs
@@ -10,7 +10,7 @@ use std::process::{Command, Stdio};
 /// builder, as we can't inspect its structure and make assertions on it
 #[derive(Debug, Default)]
 pub struct CommandBridge {
-    pub name: String,
+    pub command_name: String,
     pub args: Vec<OsString>,
     pub env: HashMap<OsString, OsString>,
     pub stdout: Option<Stdio>,
@@ -18,21 +18,21 @@ pub struct CommandBridge {
 }
 
 impl CommandBridge {
-    pub fn new(name: &str) -> Self {
+    pub fn new(command_name: &str) -> Self {
         CommandBridge {
-            name: String::from(name),
+            command_name: String::from(command_name),
             .. Default::default()
         }
     }
 
-    pub fn arg<S: AsRef<OsStr>>(&mut self, arg: S) -> &mut Self {
+    pub fn arg<S: AsRef<OsStr>>(mut self, arg: S) -> Self {
         let mut os_string = OsString::new();
         os_string.push(arg);
-        self.args.push(os_string);
+        self.args.push(os_string.to_os_string());
         self
     }
 
-    pub fn env<S: AsRef<OsStr>>(&mut self, key: S, value: S) -> &mut Self {
+    pub fn env<S: AsRef<OsStr>>(mut self, key: S, value: S) -> Self {
         let mut key_str = OsString::new();
         let mut value_str = OsString::new();
 
@@ -44,18 +44,18 @@ impl CommandBridge {
         self
     }
 
-    pub fn stderr<S: Into<Stdio>>(&mut self, stderr: S) -> &mut Self {
+    pub fn stderr<S: Into<Stdio>>(mut self, stderr: S) -> Self {
         self.stderr = Some(stderr.into());
         self
     }
 
-    pub fn stdout<S: Into<Stdio>>(&mut self, stdout: S) -> &mut Self {
+    pub fn stdout<S: Into<Stdio>>(mut self, stdout: S) -> Self {
         self.stdout = Some(stdout.into());
         self
     }
 
     pub fn to_command(self) -> Command {
-        let mut command = Command::new(self.name);
+        let mut command = Command::new(self.command_name);
         command.args(self.args);
         command.envs(self.env);
 
@@ -82,9 +82,9 @@ mod tests {
         fn it_exists() {
             // arrange
             let arg = OsString::from("some arg");
-            let mut cmd = CommandBridge::new("some cmd");
             // act
-            cmd.arg(&arg);
+            let cmd = CommandBridge::new("some cmd")
+                .arg(&arg);
             // assert
             assert!(cmd.args.contains(&arg))
         }
@@ -98,9 +98,9 @@ mod tests {
             // arrange
             let key = OsString::from("hello");
             let value = OsString::from("world");
-            let mut cmd = CommandBridge::new("plop");
             // act
-            cmd.env(&key, &value);
+            let cmd = CommandBridge::new("plop")
+                .env(&key, &value);
             // assert
             assert_eq!(&value, cmd.env.get(&key).unwrap())
         }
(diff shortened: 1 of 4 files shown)


mackwic:
I am currently is quite puzzled by the compilation error if I uncomment the line 35 of `cargo/commands.rs`. I'll find a way to fix that quickly

src/cargo/command_bridge.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    pub fn env<S: AsRef<OsStr>>(&mut self, key: S, value: S) -> &mut Self {
+        let mut key_str = OsString::new();
+        let mut value_str = OsString::new();
+
+        key_str.push(key.as_ref());
+        value_str.push(value.as_ref());
+
+        self.env.insert(key_str, value_str);
```
mgattozzi: You can just use [`to_os_string`]( https://doc.rust-lang.org/std/ffi/struct.OsStr.html#method.to_os_string)

src/cargo/command_bridge.rs
```diff
@@ -0,0 +1,141 @@
...
+        CommandBridge {
+            name: String::from(name),
+            .. Default::default()
+        }
+    }
+
+    pub fn arg<S: AsRef<OsStr>>(&mut self, arg: S) -> &mut Self {
+        let mut os_string = OsString::new();
+        os_string.push(arg);
+        self.args.push(os_string);
```
mgattozzi: Again `to_os_string` would be better. Since you keep turning them into OsStrings why not just accept something like `Into<OsString>` and calling `into`?

src/cargo/command_bridge.rs
```diff
@@ -0,0 +1,141 @@
...
+    pub env: HashMap<OsString, OsString>,
+    pub stdout: Option<Stdio>,
+    pub stderr: Option<Stdio>,
+}
+
+impl CommandBridge {
+    pub fn new(name: &str) -> Self {
+        CommandBridge {
+            name: String::from(name),
+            .. Default::default()
```
mgattozzi: This is such a neat trick!

src/cargo/command_bridge.rs
```diff
@@ -0,0 +1,141 @@
...
+
+        key_str.push(key.as_ref());
+        value_str.push(value.as_ref());
+
+        self.env.insert(key_str, value_str);
+        
+        self
+    }
+
+    pub fn stderr<S: Into<Stdio>>(&mut self, stderr: S) -> &mut Self {
```
mgattozzi: Is this supposed to be `Into<Stderr>`?

src/cargo/command_bridge.rs
```diff
@@ -0,0 +1,141 @@
...
+        self
+    }
+
+    pub fn stdout<S: Into<Stdio>>(&mut self, stdout: S) -> &mut Self {
+        self.stdout = Some(stdout.into());
+        self
+    }
+
+    pub fn to_command(self) -> Command {
+        let mut command = Command::new(self.name);
```
mgattozzi: I don't think `name` is the best choice to describe this. I think `command` would be better but that looks kind of weird here.

src/cargo/command_bridge.rs
```diff
@@ -0,0 +1,141 @@
...
+        #[test]
+        fn it_exists() {
+            // arrange
+            let mut cmd = CommandBridge::new("plop");
+            let stderr = Stdio::null();
+            // act
+            cmd.stderr(stderr);
+            // assert
+            // FIXME: is there is really no way to compare an Stdio instance ??
+            //assert_eq!(stderr as *const (), cmd.stderr.unwrap() as *const ())
```
mgattozzi: You can turn them into RawFd which is basically just a c_int and compare them.
https://doc.rust-lang.org/std/os/unix/io/type.RawFd.html

mgattozzi: Since it implements [this](https://doc.rust-lang.org/std/os/unix/io/trait.FromRawFd.html#tymethod.from_raw_fd) it also implies that it implements the into version of it.

src/cargo/commands.rs
```diff
@@ -0,0 +1,168 @@
+
+use cargo::command_bridge::CommandBridge;
+use std::path::{Path, PathBuf};
+use std::process::Stdio;
+
+use error::*;
+
+pub fn retrieve_metadata(manifest_path: &Path) -> CommandBridge {
+    let mut command = CommandBridge::new("cargo");
+    command.arg("metadata")
```
mgattozzi: I'd prefer it to be a builder pattern stylistically so you can do:

```rust
CommandBridge::new("cargo")
    .arg("metadata")
    .arg("--manifest-path")
    .arg(manifest_path)
    .arg("--no-deps")
    .arg("--format-version")
    .arg("1")
```

If you change the function `arg` to  `arg(self) -> Self` this enables that pattern. Builder patterns are great and this is essentially what you're doing here! :D

src/cargo/commands.rs
```diff
@@ -0,0 +1,168 @@
...
+        //.env("CARGO_TARGET_DIR", target_dir) // FIXME compiles
+        .env("RUSTFLAGS", "-Z save-analysis")
+        .stderr(Stdio::piped())
+        .stdout(Stdio::null());
+
+    if is_verbose {
+        command.arg("--verbose");
+    }
+
+    //match target.kind {
```
mgattozzi: Did you take this out?

src/cargo/commands.rs
```diff
@@ -0,0 +1,168 @@
...
+        .arg("check")
+        .arg("--manifest-path")
+        .arg(&manifest_path)
+        //.env("CARGO_TARGET_DIR", target_dir) // FIXME compiles
+        .env("RUSTFLAGS", "-Z save-analysis")
+        .stderr(Stdio::piped())
+        .stdout(Stdio::null());
+
+    if is_verbose {
+        command.arg("--verbose");
```
mgattozzi: Hmmm this would cause command to be consumed if you did what I mentioned. You'd have to do a let command = command.arg("--verbose");` here

src/cargo/commands.rs
```diff
@@ -0,0 +1,168 @@
...
+        fn it_returns_err_if_filename_is_not_cargo_toml() {
+            // arrange
+            let path = PathBuf::from("/users/rusty/Rusty.toml");
+            // act
+            let res = check_manifest_path_points_to_cargo_toml(&path);
+            // assert
+            assert!(res.is_err())
+        }
+    }
+}
```
mgattozzi: Is this whole file new because it got moved?

src/cargo/command_bridge.rs
mackwic: Ah ! Now that you say it, yes it is. Very useful to replace factories in tests: implement `Default` and only specify the field you want to test.

src/cargo/command_bridge.rs
mackwic: You are probably right. I was mirroring the interface of `std::process::Command` without thinking of making my life easier.

I'll test that.

src/cargo/command_bridge.rs
mackwic: ooh, I missed that. I am not familiar with the `OsStr(ing)?` interface

src/cargo/command_bridge.rs
mackwic: honestly I just mirrored the std::process::Command interface. I imagine `Stdio` is a reference to the libc streams from `stdio.h`. I didn't took the time to look for `Into<Stdio>` implementors as I need to stay compatible with std::process::Command anyway.

If you have any suggestion, I'll take it

src/cargo/command_bridge.rs
mackwic: good point. `command_name` it is

src/cargo/command_bridge.rs
mackwic: Oh ! Good catch ! Thank you ! :)

src/cargo/commands.rs
mackwic: Yes, it would be more canonic. I'll change

src/cargo/commands.rs
mackwic: 👍

src/cargo/commands.rs
mackwic: still in the WIP :)

src/cargo/commands.rs
mackwic: nope, the file is new. I extracted the logic of command building here and I want to unit test it

src/cargo/command_bridge.rs
mgattozzi: I'll be leveraging this in some of my own stuff now!

src/cargo/command_bridge.rs
mgattozzi: People tend to miss it. The Standard Library docs are great! Usually I'll take a look before using something to see what's available, especially if I tend to do common patterns like conversions!

src/cargo/command_bridge.rs
mgattozzi: I looked at the docs. I think it's Stdio as the type so this is fine, but this also means you can put an StdIn where there should be an StdErr.

src/cargo/command_bridge.rs
mgattozzi: Awesome thanks! :D

src/cargo/command_bridge.rs
mgattozzi: You're welcome :D

src/cargo/commands.rs
mgattozzi: :D

src/cargo/commands.rs
mgattozzi: Ah got it cool

src/cargo/commands.rs
mgattozzi: Even better. I'm all for more tests. Looking forward to seeing the end result :D

src/cargo/command_bridge.rs
steveklabnik: that's in std::os::unix though, what about windows?

src/cargo/command_bridge.rs
```diff
@@ -0,0 +1,141 @@
...
+use std::collections::HashMap;
+use std::convert::{AsRef, Into};
+use std::default::Default;
+use std::ffi::{OsStr, OsString};
+use std::process::{Command, Stdio};
+
+/// This temporary struct is used to inspect the generated command before calling the process::Command
+/// builder, as we can't inspect its structure and make assertions on it
+#[derive(Debug, Default)]
+pub struct CommandBridge {
```
euclio: Just another use case for https://github.com/rust-lang/rust/issues/44434 😞

mgattozzi: :( If only. Might need an RFC?

src/cargo/command_bridge.rs
mackwic: Yes this is legal. In this case on Linux it's equivalent to passing it `/dev/null` as `stdin` is _often_ closed for writing (not always though, in this case it's a blocking read for eternity)

src/cargo/command_bridge.rs
mackwic: Yes, this is much needed. And `std::{fs, io}` stubbing. So much of that

mgattozzi:
Hmmm tests are failing but it doesn't seem to be your fault (I think). These changes look great! :D

steveklabnik:
I like this idea in general, but yeah, the tests seem to be failing. do they pass locally? Given that both travis and appveyor fail with the same message, I suspect that it's something in here...

mackwic:
Sorry, not much time to work on it recently. I am still up to finish this PR, but I would understand if you close it. I can always work on it offline and re-submit it later, rebased from master.

steveklabnik:
It happens! I go for a long time without having time too.

I am going to give this a close though, if you or anyone else wants to update it, please submit a new PR! Thanks :)

Issue #98, opened 2017-08-01
Source test suite should use something more powerful than JSON pointer


euclio:
I wrote this originally as a proof of concept. What we really need is a way to query the JSON. For example, the current rustdoc uses XPath, which is a very powerful query language. The current tests are fragile because every directive depends on the exact order of the JSON. If there's more than one def kind or multiple defs in the same file, things can get messy fast.

I'm not too familiar with this domain, but there are solutions like JPATH and json-query. I'm even less familiar with the options available to Rust.

We should address this before we get too many tests :)

steveklabnik:
Marking this as a quest issue; I don't know the best solution either!

euclio:
Some options:

- JSONpath
- JMESpath (seems to have a well-tested crate!)

steveklabnik:
Marking this as hard since there's a lot of work to do; first to come up with the tech we should use, second to implement it.

euclio:
I have a proof of concept using JMESPath instead of JSON pointer. Here's what I'm thinking design-wise:

The `@has` directive takes one or two arguments. The first argument is a JMESPath, and the second argument is a JSON data type. If the two arguments evaluate to the same value, the test passes.

This means that a simple check could be

```rust
// @has included[0].type '"module"'
```

If the second argument is not supplied, it's assumed to be `true`. To check that a string contains another:

```rust
// @has 'data.attributes.docs | contains(@, "crate")'
// or
// @has 'contains(included[0].type, "crate")'
```

You can also provide a JSON object on the right side. This can be used to match more complex queries.

```rust
// @has data.relationships.modules.data[0] '{ "type": "module", "id": "modules::module" }'
```

While we no longer would have the ability to Regex match on the right side, JMESPath is extremely powerful, and I believe that `contains` would serve most of the use case. If it became a problem we could customize the JMESPath runtime with a `regex_match` function.
Thoughts, @steveklabnik?

steveklabnik:
My thoughts are mostly, "I've never heard of JMESPath before and it looks interesting, but that kinda gives me pause."

Hm.

euclio:
I hadn't heard of it either, but it seems like we are constrained by which
crates are available. The jmespath.rs crate at least fully passes the
JMESPath conformance test suite.

euclio:
We could consider using [jq](https://stedolan.github.io/jq/) as a library. That's a bit more tried and true, but we'd be introducing a C dependency.

euclio:
[...]

PR #194, opened 2017-10-18
don't run rustfmt for tests

steveklabnik:
related to https://github.com/steveklabnik/rustdoc/issues/98, our current
tests mean that some stuff has really long lines

Thoughts on this, everyone? I'm not sure there's a good way to tell rustfmt to ignore only these line lengths...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ before_install:
 script:
   - rustfmt --write-mode diff build.rs
   - rustfmt --write-mode diff $(find src -type f | grep .rs)
-  - rustfmt $(find tests -type f | grep .rs) || true
+    #- rustfmt $(find tests -type f | grep .rs) || true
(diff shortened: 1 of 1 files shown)


euclio:
Should we instead focus on splitting those long lines? Allowing them would hurt readability.

steveklabnik:
> Should we instead focus on splitting those long lines? Allowing them would hurt readability.

Allowing them would hurt?

euclio:
I was thinking that I'd rather see

```
// @has '/some/really/long/path' \
//     "Some really long documentation that really shouldn't be on one line"
```

```
// @has '/some/really/long/path' "Some really long documentation that really shouldn't be on one line"
```

and rustfmt could enforce that.

steveklabnik:
Sure, let's do that. Any chance you want to implement it? You know the test framework better than i do.

euclio:
Yeah, I can do it.

PR #202, opened 2017-10-25
Error-chain -> failure

steveklabnik:
This PR ports us from error-chain to @withoutboats' `failure` crate.

I'm done for the day so there's probably CI failures and such, but I wanted to toss this up there so that everyone can see.

/cc @aturon @alexcrichton

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,7 +4,9 @@
 #![warn(missing_docs)]
 
 #[macro_use]
-extern crate error_chain;
+extern crate failure;
+#[macro_use]
+extern crate failure_derive;
 #[macro_use]
 extern crate serde_derive;
 #[cfg_attr(test, macro_use)]
@@ -38,14 +40,19 @@ use std::path::{Path, PathBuf};
 use std::process::{self, Command, Stdio};
 
 use cargo::Target;
-use error::*;
 use json::Documentation;
 use test::TestResult;
 use ui::Ui;
 
-pub use error::{Error, ErrorKind};
 pub use json::create_documentation;
 pub use ui::Verbosity;
+
+use failure::Error;
+use failure::Fail;
+/// The general Result type for `rustdoc`.
+///
+/// Similar to `std::io::Result` and friends, this makes error handling a bit more succinct.
+pub type Result<T> = std::result::Result<T, Error>;
 
 /// A structure that contains various fields that hold data in order to generate doc output.
 #[derive(Debug)]
@@ -71,7 +78,9 @@ impl Config {
         let host = analysis::AnalysisHost::new(analysis::Target::Debug);
 
         if !manifest_path.is_file() || !manifest_path.ends_with("Cargo.toml") {
-            bail!("The --manifest-path must be a path to a Cargo.toml file");
+            return Err(failure::err_msg(
+                "The --manifest-path must be a path to a Cargo.toml file",
+            ));
         }
 
         Ok(Config {
@@ -172,9 +181,10 @@ pub fn build(config: &Config, artifacts: &[&str]) -> Result<()> {
             .stderr(Stdio::piped())
             .spawn()
             .map_err(|e| if e.kind() == io::ErrorKind::NotFound {
-                Error::with_chain(e, format!("frontend `{}` not found", frontend))
+                e.context(format!("frontend `{}` not found", frontend))
+                    .into()
             } else {
-                e.into()
+                failure::Error::from(e)
             })?;
 
         {
@@ -187,7 +197,10 @@ pub fn build(config: &Config, artifacts: &[&str]) -> Result<()> {
             task.error();
             drop(task);
             println!("\n{}", String::from_utf8_lossy(&output.stderr));
-            bail!("frontend `{}` did not execute successfully", frontend);
+            return Err(format_err!(
+                "frontend `{}` did not execute successfully",
+                frontend
+            ));
         }
     }
 
@@ -196,9 +209,9 @@ pub fn build(config: &Config, artifacts: &[&str]) -> Result<()> {
 
 /// Run all documentation tests.
 pub fn test(config: &Config) -> Result<()> {
-    let doc_json = File::open(config.documentation_path()).chain_err(
-        || "could not find generated documentation",
-    )?;
+    let doc_json = File::open(config.documentation_path()).map_err(|e| {
+        failure::Error::from(e.context("could not find generated documentation"))
+    })?;
     let docs: Documentation = serde_json::from_reader(doc_json)?;
 
     let krate = docs.data.as_ref().unwrap();
--- a/tests/source.rs
+++ b/tests/source.rs
@@ -44,6 +44,7 @@ pub type Result<T> = std::result::Result<T, Error>;
 #[fail(display = "Directive is not valid: {}", d)]
 struct InvalidDirective {
     d: String,
+    error: failure::Error,
 }
 
 #[derive(Debug, Fail)]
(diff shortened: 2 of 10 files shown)


mgattozzi:
@steveklabnik curious is there a reason to move towards that over error-chain?

src/lib.rs
```diff
@@ -172,9 +181,9 @@ pub fn build(config: &Config, artifacts: &[&str]) -> Result<()> {
             .stderr(Stdio::piped())
             .spawn()
             .map_err(|e| if e.kind() == io::ErrorKind::NotFound {
-                Error::with_chain(e, format!("frontend `{}` not found", frontend))
+                format_err!("frontend `{}` not found", frontend).into()
```
euclio: It looks like you could use `chain()` to keep the context here?

src/lib.rs
```diff
@@ -196,8 +205,8 @@ pub fn build(config: &Config, artifacts: &[&str]) -> Result<()> {
 
 /// Run all documentation tests.
 pub fn test(config: &Config) -> Result<()> {
-    let doc_json = File::open(config.documentation_path()).chain_err(
-        || "could not find generated documentation",
+    let doc_json = File::open(config.documentation_path()).map_err(|_|
```
euclio: Same here.

euclio:
This looks interesting!

steveklabnik:
> curious is there a reason to move towards that over error-chain?

It's speculated that this kind of library may be preferable to error-chain overall; I wanted to give it a shot to see what the diff looks like. I think I personally prefer it, but the only way to find out is to give it a try.

mgattozzi:
Awesome!

src/lib.rs
steveklabnik: wait, what's chain()? I don't see it in failure's docs

euclio: Looks like it's been renamed to `context`? It works slightly differently that `chain_err` did in `error_chain`.

withoutboats: Yea I renamed it to context. You could do `e.context(format!(...)).into()` here.

steveklabnik:
@withoutboats while working on the test part of this PR, 

```
    = note: expected type `std::result::Result<_, failure::Error>`
               found type `std::result::Result<_, failure::Context<InvalidDirective>>`
```

Do you have to do something specific to use `Context`? It doesn't implement `Error`?

withoutboats:
`Context` implements `Fail`; to get an `Error` all `Fail`ures have to be cast using `?` or `into()`.

steveklabnik:
Thanks boats!

Now our test is failing because of the way we test the context; I suspect that I have to update the macro in some way. More work needed.

euclio:
In this case, it might just be better to check that the cause is correct and always return an `InvalidDirective` failure from `parse_test`?

steveklabnik:
I've rebased this, and it compiles, but has test failures about not being able to find the source test crates. Hrm.

steveklabnik:
Ohh interesting! https://travis-ci.org/steveklabnik/rustdoc/jobs/305812561

even though https://github.com/steveklabnik/rustdoc/pull/203 passed, it's now failing. So it's likely *that* PR, not this one.... @euclio any ideas?

euclio:
Failing on master too. Probably related to a change in how the analysis is generated. I did a quick test and it should be fixable by updating `rls-analysis` and `rls-data`, though there's an unrelated failure in `structs.rs` as well.

steveklabnik:
ah, that makes perfect sense.

ill dig in.

steveklabnik:
Updated and rebased, still failing. I must have missed something somewhere...

steveklabnik:
oh wait looks like a bad rebase....

steveklabnik:
Got master passing, this seems to reintroduce it. Hmmmmmmmmm

steveklabnik:
@mgattozzi @euclio okay I think this is ready for a review, maybe you'll spot where I messed up. @withoutboats if you have a second as well, that'd be nice.

tests/source.rs
```diff
@@ -401,37 +423,49 @@ fn parse_test(line: &str) -> Option<Result<TestCase>> {
...
-            let regex = Regex::new(&args[1]).chain_err(|| "invalid regex syntax")?;
+            let regex = Regex::new(&args[1]).map_err(|e| {
+                failure::Error::from(e.context("invalid regex syntax"))
+            })?;
 
             Directive::Has(regex)
         }
         "matches" => {
-            ensure!(args.len() == 2, "not enough arguments");
+            if args.len() == 2 {
```
euclio: This should be `if args.len() != 2`

tests/source.rs
```diff
@@ -401,37 +423,49 @@ fn parse_test(line: &str) -> Option<Result<TestCase>> {
...
-            )?;
+            let value = serde_json::from_str(&args[1]).map_err(|e| {
+                failure::Error::from(e.context("invalid JSON syntax"))
+            })?;
 
             Directive::Matches(value)
         }
         "assert" => {
-            ensure!(args.len() == 1, "expected 1 argument");
+            if args.len() == 1 {
```
euclio: This should be `if args.len() != 1`

tests/source.rs
steveklabnik: lol oops

steveklabnik:
Awesome, thank you! Of course I messed that up... that leads to just one failure now.

tests/source.rs
```diff
@@ -391,7 +413,7 @@ fn parse_test(line: &str) -> Option<Result<TestCase>> {
     if let Some(caps) = DIRECTIVE_RE.captures(line) {
         let directive = &caps["directive"];
         let result = parse_directive(directive, &caps["args"], caps.name("negated").is_some())
-            .chain_err(|| ErrorKind::InvalidDirective(line.into()));
+            .map_err(|e| e.context(InvalidDirective { d: line.into() }).into());
```
euclio: I think the context is backwards here. I think that `parse_test` should return `Option<Result<TestCase, InvalidDirective>>`. Make sure to set the context of the `InvalidDirective` error to be the original error. The tests should then check that `err.cause()` is the expected type.

steveklabnik: yeah, this is sorta what i'm thinking as well. been toying with this...

tests/source.rs
```diff
@@ -401,37 +423,49 @@ fn parse_test(line: &str) -> Option<Result<TestCase>> {
 fn parse_directive(directive: &str, args: &str, negated: bool) -> Result<TestCase> {
     let args = match shlex::split(args) {
```
euclio: This can be `ok_or_else`

euclio:
Looks good with comments addressed. Any reason you're preferring `ok_or` over `ok_or_else`? I don't think clippy likes that.

steveklabnik:
> Any reason you're preferring ok_or over ok_or_else? I don't think clippy likes that.

no reason; I'll run clippy too. 

now dealing with this fun error:

```
        thread 'tests::parse_test' panicked at 'unexpected error kind, got InvalidDirective { d: "// @has /test \'[\'" }

invalid JMESPath syntax but expected "InvalidDirective"', tests\source.rs:583:8
```

hmmmmmmmmmmmmmmmmm

euclio:
Yeah, I was confused by that too. Does downcasting the cause to `jmespath::JmespathError` work?

steveklabnik:
> Yeah, I was confused by that too. Does downcasting the cause to jmespath::JmespathError work?

It doesn't seem to, no matter what I try to cast it to, it doesn't seem to do the right thing. Hrm.

steveklabnik:
@euclio so check out this last commit: if we do make it always return `InvalidDirective`, then there's no reason to assert that type about it. this is now doing something similar to https://boats.gitlab.io/failure/error-errorkind.html; maybe we should do that exactly?

euclio:
Exactly. Error/ErrorKind could work, but since this is an internal-only error that complexity might not be warranted. We could just make it an enum like [here](https://boats.gitlab.io/failure/custom-fail.html#caveats-on-this-pattern).

steveklabnik:
Time to :shipit: , if we want to tweak that last error, we can, but I don't think it matters a ton, honestly.

PR #208, opened 2017-11-26
Allow binary to be used from another directory

hjr3:
This change moves the indoc crate to be a dev-dependency. If the indoc
crate is included in the binary, then it will depend on the proc_macro
shared library provided by rust.


steveklabnik:
merged manually as https://github.com/steveklabnik/rustdoc/commit/dd9af9da5466d01982b1c6327faa55aa207ed186 to fix the conflicts, thank you!

PR #210, opened 2017-11-28
Modules: parent links

steveklabnik:
We're gonna need to modify the JSON slightly to add parent links for modules. In theory, libraries could figure out that this is true, but in reality, they don't, and so this gives them a bit less work to do.

I have a failing test here, but am not likely to get it passing until tomorrow. If anyone wants to take a shot at this in the meantime, it would be very helpful!

### DIFF
--- a/tests/source/nested_modules.rs
+++ b/tests/source/nested_modules.rs
@@ -1,20 +1,46 @@
 #![crate_type = "lib"]
 
+// Testing the child information:
+//
 // @has "included[?id=='nested_modules::example_module'] \
 //          .relationships.modules.data[].type | [0]" \
 //      "module"
 // @has "included[?id=='nested_modules::example_module'] \
 //          .relationships.modules.data[].id | [0]" \
 //      "nested_modules::example_module::nested"
+//
+// Testing that we don't have a parent:
+//
+// @matches "included[?id=='nested_modules::example_module'] \
+//          .relationships.parent" []
 pub mod example_module {
 
+    // For the child:
+    //
     // @has "included[?id=='nested_modules::example_module::nested'] \
     //          .relationships.modules.data[].type | [0]" \
     //      "module"
     // @has "included[?id=='nested_modules::example_module::nested'] \
     //          .relationships.modules.data[].id | [0]" \
     //      "nested_modules::example_module::nested::nested2"
+    //
+    // For the parent:
+    //
+    // @has "included[?id=='nested_modules::example_module::nested'] \
+    //          .relationships.parent.data.type | [0]" \
+    //      "module"
+    // @has "included[?id=='nested_modules::example_module::nested'] \
+    //          .relationships.parent.data.id | [0]" \
+    //      "nested_modules::example_module"
     pub mod nested {
+        // For the parent:
+        //
+        // @has "included[?id=='nested_modules::example_module::nested::nested2'] \
+        //          .relationships.parent.data.type | [0]" \
+        //      "module"
+        // @has "included[?id=='nested_modules::example_module::nested::nested2'] \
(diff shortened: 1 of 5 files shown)


tests/source/nested_modules.rs
```diff
@@ -8,12 +10,23 @@
 //      "nested_modules::example_module::nested"
 pub mod example_module {
```
steveklabnik: @euclio i would also like to add a test above, to show that this module does *not* have a parent link. I tried to use `@!has` for this, but couldn't quite get it, any thoughts?

steveklabnik:
Ah, looks like this is actually pretty easy...

steveklabnik:
This should pass the above test, but the negative test I referred to above will fail:

```text
  "relationships": {
        "structs": {
          "data": [
            {
              "type": "struct",
              "id": "example::example_module::AnotherStruct"
            }
          ]
        },
        "parent": {
          "data": [
            {
              "type": "module",
              "id": "example::"
            }
          ]
        }
```

this makes sense, as modules at the root do have the anonymous root module as their parent.

steveklabnik:
I've fixed the bug, but still don't know how to write the test. Let's not merge till we have that so that we don't regress.

steveklabnik:
Okay, so this is also slightly off: parent shouldn't be an array, just an object.

steveklabnik:
Looks like https://docs.rs/jsonapi/0.5.2/jsonapi/api/enum.IdentifierData.html will basically be how we need to change relationships; an enum of vec or hashmap.

tests/source/nested_modules.rs
euclio: Something like

```
// @matches "included[?id=='nested_modules::example_module'] \
//          .relationships.parent" []
```

Should do the trick.

steveklabnik: ah, nice

steveklabnik:
i pushed the refactoring alone as https://github.com/steveklabnik/rustdoc/pull/210/commits/062a8b70b913f714429944c93afd992b3ba91e99 so it's easy to verify that the test changes i made are good

steveklabnik:
Hm, I expected this test to now pass, because it's generating it correctly: (this is for the example crate)

```json
    {
      "type": "module",
      "id": "example::nested1::nested2",
      "attributes": {
        "name": "nested2",
        "docs": " nested 2\n"
      },
      "relationships": {
        "parent": {
          "data": {
            "type": "module",
            "id": "example::nested1"
          }
        },
        "modules": {
          "data": [
            {
              "type": "module",
              "id": "example::nested1::nested2::nested3"
            }
          ]
        }
      }
```

more investigation needed but we're close!

steveklabnik:
```text
        error: test failed on line 29
caused by: test failed on line 29
backtrace, if any:
thread 'docs::nested_modules' panicked at 'explicit panic', C:\Users\steve\src\rustdoc\target\debug\build\rustdoc-b0104c9e001a8e5b\out/source_generated.rs:1:3750
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

but 

```json
   {
      "type": "module",
      "id": "example::example_module::nested",
      "attributes": {
        "name": "nested",
        "docs": ""
      },
      "relationships": {
        "parent": {
          "data": {
            "type": "module",
          }
        },
        "modules": {
          "data": [
            {
              "type": "module",
              "id": "example::example_module::nested::nested2"
            }
          ]
        }
      }
    },
```

.... do I have the test syntax wrong or something?

steveklabnik:
Yup, I broke the tests when I "fixed" them. Sigh.

This should now pass!

steveklabnik:
apparently the os x builders are.... acting strange

PR #215, opened 2017-12-03
Create flag-compatible infrastructure

hjr3:
Adds infrastructure to become flag-compatible with existing rustdoc.
There are a few general buckets a flag falls into:

1. Flag stays the same
2. Flag is no longer supported in the new rustdoc
3. Flag has moved (usually under a subcommand)

Most flags have still not been implemented and they will say so. Also,
some flags are in a limbo state where they are probably going to be
unsupported in the future. Until that decision has been declared, they
are marked as unimplemented.

BREAKING: This change moves `-o` from open flag to output flag.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -26,3 +26,21 @@ pub struct Json {
     /// The location of the unexpected JSON
     pub location: String,
 }
+
+/// Thrown when a flag, supported in the original rustdoc, is no longer supported
+#[derive(Debug, Fail)]
+#[fail(display = "Unsupported flag: {}", flag_name)]
+pub struct UnsupportedFlag {
+    /// The name of the unsupported flag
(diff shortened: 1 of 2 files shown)


hjr3:
I spent some time thinking of a way to not have to declare the unsupported options and then have a separate list to check against them again. Everything I came up with added too much complexity to be worth it. If someone has an idea, I am all for it.

steveklabnik:
Thanks! I can't give a real review until tomorrow, but this is failing due to lack of `rustfmt`.

hjr3:
Hmm, I did not think the long strings would fail rustfmt. If I wrap them, it messes up the usage output. I will figure this out in a bit.

steveklabnik:
I re-did the formatting for you. Thanks!

PR #222, opened 2017-12-31
redo rustfmt

steveklabnik:
rustfmt is a component now, and actually handles `build.rs` and tests and such.

We also need to run it, as we were previously using the old, stable rustfmt, and some stuff has changed.


steveklabnik:
traivs timed out, it's fine.

PR #224, opened 2018-01-06
Fix failing CI due to rustfmt regression

### LINKED ISSUE #223: CI failing due to rustfmt update
In b5c3ec5c2d3bf4862fec4e2b3f41a286b913804c a lot of `json!()` macro bodies were removed, which caused tests to fail. This change was done because 5d3f317a74297ac836136378dedae350a39c1ff7 updated to use the new rustfmt. Unfortunately, there is a regression, https://github.com/rust-lang-nursery/rustfmt/issues/2344, where the new rustfmt removes macro bodies when the macro is assigned in a let statement.

hjr3:
This fixes the failing tests and adds a temporary fix for a regression in rustfmt so the CI process
will correctly run. I broke this into two commits so (theoretically) we can just back out the temporary fix.

Fixes #223

### DIFF
--- a/src/cargo.rs
+++ b/src/cargo.rs
@@ -248,6 +248,8 @@ mod tests {
     fn target_from_metadata() {
         let ui = Ui::default();
 
+        // work around until https://github.com/rust-lang-nursery/rustfmt/issues/2344 is fixed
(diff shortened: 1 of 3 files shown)


hjr3:
I broke this into two commits so (theoretically) we can just back out the temporary fix.

PR #225, opened 2018-01-06
Do not check subcommands for moved flags

hjr3:
In b864a9e, the check for moved flags was using `is_present` which
checks flags in subcommands. This caused a lot of false positives.
Instead, look at the args at the top level only and check if the flag
exists there.


hjr3:
This will fail CI until https://github.com/steveklabnik/rustdoc/pull/224 is merged

hjr3:
Rebased and passing CI now.

